PR #5, opened 2022-11-14
Updating data model

southeo:
Updating data model: separating request and response objects, and simplifying both.


src/test/java/com/example/handlemanager/service/PidTypeServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+	private HandleRepository handleRep;
+	
+	@Mock
+	private Clock clock;
+	
+	@InjectMocks
+	private PidTypeService pidTypeService;
+	
+	// Time stuff
+	private MockedStatic mockedStatic;
```
samleeflang, line 38: Don't think it is used here

src/test/java/com/example/handlemanager/service/PidTypeServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+	private long timestamp;
+	
+	// These are what the values are called in the handle record
+	private String pid;
```
samleeflang, line 42: Don't think these need to be uninitiated private class variables

src/test/java/com/example/handlemanager/service/PidTypeServiceTest.java
```diff
@@ -0,0 +1,129 @@
...
+		
+		if (isDoi) {
+			registrationAgencyDoiName = PTR_REGISTRATION_DOI_NAME;
+			record.add(new Handles(recordPid, i++, "registrationAgencyDoiName", registrationAgencyDoiName, timestamp));
+		}
+		return record;
+	}
+	
+	private long initTimestamp() {
+		//CREATED_INSTANT.getEpochSecond();
```
samleeflang, line 125: Remove comment

samleeflang (commented):
Hi Sou,
Nice work, there are a lot of things in there that we need.
It is your first reviewed PR so part of the comments are about conventions.
It also reminds me that we haven't written these down which might be a good idea.
I will plan an hour tomorrow where we can run through them in person and see what we can do differently.
We can then also walk through the outstanding Sonar issues

pom.xml
```diff
@@ -8,22 +8,22 @@
 		<version>2.7.0</version>
 		<relativePath/> <!-- lookup parent from repository -->
 	</parent>
-
+	
 	<groupId>com.example</groupId>
```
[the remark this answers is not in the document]
southeo: So have:  `<groupId>eu.dissco</groupId>` ? Yes that makes sense

pom.xml
```diff
@@ -126,7 +159,11 @@
     		<groupId>org.glassfish.jersey.core</groupId> <!-- Response Package 2-->
```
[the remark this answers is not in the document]
southeo: I think this was related to when I introduced openFeign - not needed any more & removed

pom.xml
```diff
@@ -75,40 +83,65 @@
 		<dependency>
       		<groupId>io.github.openfeign</groupId>
       		<artifactId>feign-okhttp</artifactId>
-      	</dependency>
+      	</dependency> 
       	<dependency>
-  			<groupId>io.github.openfeign</groupId>
-   			<artifactId>feign-httpclient</artifactId>
+    		<groupId>com.h2database</groupId>
```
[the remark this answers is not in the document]
southeo: Not anymore

src/main/java/com/example/handlemanager/configuration/Config.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.time.Clock;
+
+@Configuration
+@RequiredArgsConstructor
+public class Config {
+	
+	@Bean
+	public Clock clock() {
+		return Clock.systemUTC();
```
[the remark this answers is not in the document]
southeo: I think the default is actually `Clock.systemDefaultZone()`

src/main/java/com/example/handlemanager/domain/requests/DigitalSpecimenBotanyRequest.java
```diff
@@ -0,0 +1,46 @@
+package com.example.handlemanager.domain.requests;
+
+import lombok.Data;
+
+@Data
+public class DigitalSpecimenBotanyRequest extends DigitalSpecimenRequest {
+	final String objectType;
```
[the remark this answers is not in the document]
southeo: @value makes the class itself final. There's a lot of inheritance with these request classes, so that's not ideal. I'll just go with @Getter for now and see if I need more lombok functionality.

src/main/java/com/example/handlemanager/repository/HandleRepository.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+
+@Repository
+public interface HandleRepository extends JpaRepository<Handles, HandleIdx> {
+	// Get all handles	
+	@Query(value="select distinct handle from handles h", nativeQuery=true)
```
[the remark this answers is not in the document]
southeo: Good catch, added basic pagination for now (will change when jooq is implemented though)

src/main/java/com/example/handlemanager/service/PidTypeService.java
```diff
@@ -0,0 +1,76 @@
...
+
+	@Autowired
+	public HandleRepository handleRep;
+	
+	public String resolveTypePid(String typePid) {
+		String pidRecord = "";
+		try {
+			pidRecord = resolveTypePidVal(typePid);
+		} catch (PidResolutionException e) {
+			e.printStackTrace();
```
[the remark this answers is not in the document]
southeo: Handled by controller now

src/main/java/com/example/handlemanager/service/PidTypeService.java
```diff
@@ -0,0 +1,76 @@
...
+		List<Handles> typeRecord = handleRep.resolveHandle(typePid.getBytes());
+		if (typeRecord.isEmpty()) {
+			throw new PidResolutionException("Unable to resolve type PID");
+		}
+
+		String pid = getDataFromType("pid", typeRecord);
+		String primaryNameFromPid = getDataFromType("primaryNameFromPid", typeRecord);
+		String pidType;
+		String registrationAgencyDoiName = "";
+		String typeJson = "";
```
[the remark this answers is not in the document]
southeo: I'm not quite sure which info you're talking about. 
There are a few fields in the pid record that are themselves json-formatted strings of pid records of types. The strings are in the following format: 
{
"pid": "10.20/ABC-123",
"pidType: "<handle or doi>"
"primaryNameFromPid":"<Institution Name>"
}

This class uses the pid of the type record provided in the request (i.e. 10.20/ABC-123) to resolve the full record and create the above string. 

Do you mean should this string be in the database? I don't think it should. 

A better way of doing this would be creating a POJO with these fields and using an object mapper to generate the string.

src/main/java/com/example/handlemanager/service/PidTypeService.java
```diff
@@ -0,0 +1,76 @@
...
+		final String NEWLINE = "\", \n";
+
+		if (pid.contains("doi")) {
+			pidType = "doi";
+			registrationAgencyDoiName = getDataFromType("registrationAgencyDoiName", typeRecord);
+
+			typeJson = "{ \n" + "\"pid\": \"" + pid + NEWLINE + "\"pidType\": \"" + pidType + NEWLINE
```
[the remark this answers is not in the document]
southeo: Done

src/main/java/com/example/handlemanager/service/PidTypeService.java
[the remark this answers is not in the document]
southeo: Ignore the comment about pojo, objectNode is way better

pom.xml
```diff
@@ -1,146 +1,133 @@
...
+      <groupId>org.springframework.boot</groupId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <artifactId>jacoco-maven-plugin</artifactId>
+      <groupId>org.jacoco</groupId>
+      <version>0.8.8</version>
+    </dependency>
+  </dependencies>
+  <description>API for adding handle records</description>
```
samleeflang, line 112: Normally these go on top.
Then dependencies followed by plugins.
Is better for readability and maintainability as you first get the name, parent, and versions

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
```diff
@@ -0,0 +1,156 @@
...
+    List<String> handleList = service.getHandlesPaged(pidStatus, pageNum, pageSize);
+    if (handleList.isEmpty()) {
+      throw new PidResolutionException("Unable to resolve pids");
+    }
+    return ResponseEntity.ok(handleList);
+  }
+
+  @ExceptionHandler(PidCreationException.class)
+  private ResponseEntity<String> pidCreationException(PidCreationException e) {
+    return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.getMessage());
```
samleeflang, line 142: Maybe use another return code.
Then 400 range is for exception because there is some condition invalid and you 'expected' this exception.
The 500 range is something happend that you did not anticipate and you don't know why.
For more on the difference:
https://enterprisecraftsmanship.com/posts/rest-api-response-codes-400-vs-500/#:~:text=The%20main%20difference%20between%20the,itself%20who%20caused%20the%20problem.

src/main/java/eu/dissco/core/handlemanager/domain/requests/DigitalSpecimenBotanyRequest.java
```diff
@@ -0,0 +1,60 @@
+package eu.dissco.core.handlemanager.domain.requests;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.Objects;
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@ToString
+@JsonIgnoreProperties
```
samleeflang, line 10: Why is this needed?

src/main/java/eu/dissco/core/handlemanager/domain/requests/HandleRecordRequest.java
```diff
@@ -0,0 +1,42 @@
...
+@JsonIgnoreProperties
+public class HandleRecordRequest {
+
+  protected final String pidIssuerPid;
+  protected final String digitalObjectTypePid;
+  protected final String digitalObjectSubtypePid;
+  protected final String[] locations;
+
+  @Override
+  public boolean equals(Object o) {
```
samleeflang, line 22: This doesn't inherit any method so I think the lombok annotation can be used, although for consistency it might be good  to keep the methods in full

src/main/java/eu/dissco/core/handlemanager/domain/requests/HandleRecordRequest.java
```diff
@@ -0,0 +1,42 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+
+@Getter
+@RequiredArgsConstructor
+@ToString
+@JsonIgnoreProperties
+public class HandleRecordRequest {
+
+  protected final String pidIssuerPid;
```
samleeflang, line 16: Can these be private final variables?

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenBotanyResponse.java
```diff
@@ -0,0 +1,27 @@
+package eu.dissco.core.handlemanager.domain.responses;
+
+import eu.dissco.core.handlemanager.repositoryobjects.Handles;
+import java.util.List;
+import lombok.Data;
+
+@Data
```
samleeflang, line 7: I think these can be the same as the others right? Only Getters and the constructor for init. The variables can then become private final.

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenResponse.java
```diff
@@ -0,0 +1,31 @@
+package eu.dissco.core.handlemanager.domain.responses;
+
+import eu.dissco.core.handlemanager.repositoryobjects.Handles;
+import java.util.List;
+import lombok.Data;
+
+@Data
```
samleeflang, line 7: I think these can be the same as the others right? Only Getters and the constructor for init. The variables can then become private final.

src/main/java/eu/dissco/core/handlemanager/domain/responses/DoiRecordResponse.java
```diff
@@ -0,0 +1,28 @@
+package eu.dissco.core.handlemanager.domain.responses;
+
+import eu.dissco.core.handlemanager.repositoryobjects.Handles;
+import java.util.List;
+import lombok.Data;
+
+@Data
```
samleeflang, line 7: I think these can be the same as the others right? Only Getters and the constructor for init. The variables can then become private final.

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenResponse.java
```diff
@@ -0,0 +1,31 @@
...
+public class DigitalSpecimenResponse extends DoiRecordResponse {
+
+  String digitalOrPhysical;
+  String specimenHost;
+  String inCollectionFacility;
+
+  public DigitalSpecimenResponse(List<Handles> entries) {
+    super(entries);
+    String type;
```
samleeflang, line 16: I don't get this constructor.
You are overwriting the type multiple times right?
Say entry 1 is type 'A' and entry 2 is type 'B' then the type and data is overwritten.
I have the feeling that you are creating one object while it should be many.

src/main/java/eu/dissco/core/handlemanager/domain/responses/HandleRecordResponse.java
```diff
@@ -0,0 +1,45 @@
...
+        case "pid" -> this.pid = data;
+        case "pidIssuer" -> this.pidIssuer = data;
+        case "digitalObjectType" -> this.digitalObjectType = data;
+        case "digitalObjectSubtype" -> this.digitalObjectSubtype = data;
+        case "10320/loc" -> this.locs = data;
+        case "issueDate" -> this.issueDate = data;
+        case "issueNumber" -> this.issueNumber = data;
+        case "pidKernelMetadataLicense" -> this.pidKernelMetadataLicense = data;
+        case "HS_ADMIN" -> this.hsAdmin = data;
+        default -> log.info("Base constructor called");
```
samleeflang, line 39: Not really where default is for, I would break when the case is met and use default to indicate that none of the above cases matches. The log statement (although I would probably not put one here) at the top of the constructor

src/main/java/eu/dissco/core/handlemanager/service/HandleGeneratorService.java
```diff
@@ -0,0 +1,159 @@
...
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class HandleGeneratorService {
+
+  private static final String ALPHA_NUM = "ABCDEFGHJKLMNPQRSTUVWXYZ1234567890";
+  private static final String PREFIX = "20.5000.1025/";
+  private final char[] symbols;
+  private final char[] buf;
+  @Autowired
```
samleeflang, line 26: No longer needed

src/main/java/eu/dissco/core/handlemanager/service/HandleGeneratorService.java
```diff
@@ -0,0 +1,159 @@
...
+  private HandleRepository handleRep;
+  @Autowired
+  private Random random;
+
+  public HandleGeneratorService() {
```
samleeflang, line 31: I think this could conflict with the annotation. For not hardcode these with the variable and maybe make a task to create a properties class where we can change them for env variables.

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    // Save all records
+    handleRep.saveAll(handleRecordsAll);
+    return response;
+  }
+
+  public List<DoiRecordResponse> createDoiRecordBatch(List<DoiRecordRequest> requests)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    List<byte[]> handles = hf.genHandleList(requests.size());
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> doiRecord;
```
samleeflang, line 99: Why not init this variable in the loop so that it is just instantiated there and you don't have to clear?

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    return response;
+  }
+
+  public List<DigitalSpecimenResponse> createDigitalSpecimenBatch(
+      List<DigitalSpecimenRequest> requests)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    List<byte[]> handles = hf.genHandleList(requests.size());
+
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> digitalSpecimenRecord;
```
samleeflang, line 129: Why not init this variable in the loop so that it is just instantiated there and you don't have to clear?

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    handleRep.saveAll(digitalSpecimenRecordsAll);
+    return response;
+  }
+
+  public List<DigitalSpecimenBotanyResponse> createDigitalSpecimenBotanyBatch(
+      List<DigitalSpecimenBotanyRequest> requests)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    List<byte[]> handles = hf.genHandleList(requests.size());
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> digitalSpecimenBotanyRecord;
```
samleeflang, line 158: Why not init this variable in the loop so that it is just instantiated there and you don't have to clear?

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    handleRep.saveAll(digitalSpecimenRecordsAll);
+    return response;
+  }
+
+  public HandleRecordResponse createHandleRecord(HandleRecordRequest request)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    byte[] handle = hf.genHandleList(1).get(0);
+
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> handleRecord;
```
samleeflang, line 188: Instantiation is on next line, combine statements

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    handleRecord = prepareHandleRecord(request, handle, timestamp);
+    List<Handles> posted = handleRep.saveAll(handleRecord);
+    return new HandleRecordResponse(posted);
+  }
+
+  public DoiRecordResponse createDoiRecord(DoiRecordRequest request)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    byte[] handle = hf.genHandleList(1).get(0);
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> handleRecord;
```
samleeflang, line 198: Instantiation is on next line, combine statements

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    handleRecord = prepareDoiRecord(request, handle, timestamp);
+    List<Handles> posted = handleRep.saveAll(handleRecord);
+    return new DoiRecordResponse(posted);
+  }
+
+  public DigitalSpecimenResponse createDigitalSpecimenRecord(DigitalSpecimenRequest request)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    byte[] handle = hf.genHandleList(1).get(0);
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> handleRecord;
```
samleeflang, line 208: Instantiation is on next line, combine statements

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -0,0 +1,391 @@
...
+    List<Handles> posted = handleRep.saveAll(handleRecord);
+    return new DigitalSpecimenResponse(posted);
+  }
+
+  public DigitalSpecimenBotanyResponse createDigitalSpecimenBotanyRecord(
+      DigitalSpecimenBotanyRequest request)
+      throws PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
+    byte[] handle = hf.genHandleList(1).get(0);
+    long timestamp = Instant.now().getEpochSecond();
+    List<Handles> handleRecord;
```
samleeflang, line 219: Instantiation is on next line, combine statements

src/main/java/eu/dissco/core/handlemanager/service/PidTypeService.java
```diff
@@ -0,0 +1,82 @@
...
+    return mapper.writeValueAsString(objectNode);
+  }
+
+  private String getDataFromType(String type, List<Handles> hList) {
+    for (Handles h : hList) {
+      if (h.getType().equals(type)) {
+        return h.getData();
+      }
+    }
+    return ""; // This should maybe return a warning?
```
samleeflang, line 79: Agree probably a log.error and maybe an exception but at least log that something unexpected happend

src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+  @BeforeEach
+  void setup() {
+    controller = new HandleController(service);
+    handle = TestUtils.HANDLE.getBytes();
+  }
+
+  @Test
+  void testHandleRecordCreation()
+      throws PidCreationException, PidResolutionException, JsonProcessingException, ParserConfigurationException, TransformerException {
```
samleeflang, line 49: For these test methods you could also just `throws Exception`

src/test/java/eu/dissco/core/handlemanager/service/HandleGeneratorServiceTest.java
```diff
@@ -0,0 +1,122 @@
...
+@Slf4j
+class HandleGeneratorServiceTest {
+
+  @Mock
+  private HandleRepository handleRep;
+
+  @Mock
+  private Random random;
+
+  @InjectMocks
```
samleeflang, line 29: You probably want to test with the actual service

src/test/java/eu/dissco/core/handlemanager/service/HandleServiceTest.java
```diff
@@ -0,0 +1,359 @@
...
+  private HandleService service;
+
+  private List<byte[]> handlesList;
+
+  private MockedStatic<Instant> mockedStatic;
+
+
+  @BeforeEach
+  void setup() throws PidResolutionException, JsonProcessingException {
+    log.info("tbf is null: " + String.valueOf(documentBuilderFactory == null));
```
samleeflang, line 80: Is this needed, the documentBuilderFactor is always null right? as it isn't initiated

src/test/java/eu/dissco/core/handlemanager/service/HandleServiceTest.java
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  private List<Handles> generateBatchDigitalSpecimenBotanyList() {
+    List<Handles> handleList = new ArrayList<>();
+    for (byte[] h : handlesList) {
+      handleList.addAll(generateTestDigitalSpecimenBotanyRecord(h));
+    }
+    return handleList;
+  }
+
```
samleeflang, line 342: Sometimes there are some unnecessary white lines in your code

src/test/java/eu/dissco/core/handlemanager/service/PidTypeServiceTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+@Slf4j
+class PidTypeServiceTest {
+  // NOTE: Pid Type Record => PTR in naming convention because these
+  // PTR refers to the handle record that stores information about a type
+
+  ObjectMapper mapper;
```
samleeflang, line 28: Doesn't need to be declared here

samleeflang (changes requested):
Looking good, couple of things and then this can be merged

src/main/java/eu/dissco/core/handlemanager/domain/requests/HandleRecordRequest.java
southeo: I'd like to keep them consistent

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenResponse.java
southeo: Resolving because I removed this constructor when I implemented Jooq

(To clarify: the intention was that each "entries" object have a different "type"; the list provided was the same thing as a single handle record. So one entry might be [type: "pidStatus", data: "DRAFT"], then the next would be [type: "digitalObjectType", data: "Botany Specimen"], etc.

src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
southeo: Done

src/main/java/eu/dissco/core/handlemanager/service/HandleGeneratorService.java
southeo: Fixed + added task to jira

src/main/java/eu/dissco/core/handlemanager/domain/responses/HandleRecordResponse.java
southeo: Resolved because this is removed in update

src/main/java/eu/dissco/core/handlemanager/domain/responses/DoiRecordResponse.java
southeo: Replaced all classes with getter and constructor tags

src/test/java/eu/dissco/core/handlemanager/service/HandleGeneratorServiceTest.java
southeo: Resolved because this issue is fixed in the Jooq branch

southeo (commented):
I'm removing some tests from this PR. I rewrote most of the tests to comply with Jooq in the next PR.

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
southeo: Oh I didn't know the difference between these two ranges, that's interesting. Marked as 400 BAD REQUEST.

src/main/java/eu/dissco/core/handlemanager/domain/requests/DigitalSpecimenBotanyRequest.java
southeo: I thought this prevented an UnknownFieldException, in case a field from a DoiRequest was included in a HandleRequest. But I can't seem to trigger that error now...
Removed

src/test/java/eu/dissco/core/handlemanager/service/PidTypeServiceTest.java
southeo: Are you sure? this is an injected bean.

southeo:
PR is reviewed. All changes that have been implemented have been resolved; anything outstanding is remaining (a few comments). Mostly, there are a few things that I have already fixed downstream in the JOOQ branch, so I didn't touch them in this branch. Thanks for the feedback.

src/test/java/eu/dissco/core/handlemanager/service/PidTypeServiceTest.java
samleeflang: No it is not here as this is a test class where we only Mock things but don't inject. Here it is just an uninstantiated variable. In the constructor you instantiate it an pass it to the service but in the test itself is not used. So this can be removed here and only be instantiated and used in the constructor.

samleeflang (approved):
Ok thanks for picking up the comments! Have a few left but let's do those in the JOOQ PR and I will approve this on. One additional comment is that sometimes a multiple new lines are used which doesn't break the code but for consistency just use one between methods.
- HandleController on line 152 still returns a 500. We don't throw the 500 if we expect an error which is here so better to use a 4XX
- Instead of questions or TODO in the code create a task on the backlog, so the comment on HandleService 362 can be removed. Don't think there is a converter, however this would be a place to use a stream. Something like bytelist.toStream().map(byteToString).toList() which is also optimized for this kind of actions. More on streams see:  https://stackify.com/streams-guide-java-8/
- Didn't see a log statement in the PidTypeService instead of the comment
- Did anything happen with this comment Is this needed, the documentBuilderFactor is always null right? as it isn't initiated in HandleServiceTest. It is strange to log in a test class and especially in a constructor
- Remove the ObjectMapper on line 26 of the PidTypeServiceTest

PR #11, opened 2022-12-02
Updating data model jooq

southeo:
Jooq replacing JPA as repository layer (includes Jooq-generated code)

- "Handles" object replaced with "HandleAttribute" object, more in-line with processing 
- Still need to iterate over a list of rows to build a response object (because that's how information is stored in the database). However, now the repository layer explicitly sets each attribute, so it's more reliable.


src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenBotanyResponse.java
```diff
@@ -1,27 +1,44 @@
...
 
-import eu.dissco.core.handlemanager.repositoryobjects.Handles;
-import java.util.List;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
 
 @Data
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
```
samleeflang, line 9: This makes it a very mutable object. Does this mean you instantiate it and then set all properties? It is the opposite of what we tried to do with the getters and allargs where we created immutable objects.

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenBotanyResponse.java
```diff
@@ -1,27 +1,44 @@
...
 public class DigitalSpecimenBotanyResponse extends DigitalSpecimenResponse {
 
   String objectType;
```
samleeflang, line 12: Can these be private?

src/test/java/eu/dissco/core/handlemanager/repository/HandleRepositoryIT.java
```diff
@@ -0,0 +1,263 @@
...
+
+  @Test
+  void testCreateHandle() throws PidCreationException {
+    // Given
+    byte[] handle = HANDLE.getBytes();
+    List<HandleAttribute> attributes = generateTestHandleAttributes(handle);
+    HandleRecordResponse responseExpected = generateTestHandleResponse(handle);
+
+    // When
+    HandleRecordResponse responseReceived = handleRep.createHandle(handle, CREATED, attributes);
```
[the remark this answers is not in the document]
southeo: Gotcha, I'll keep this in mind moving forward. Do you have specific changes you want with this?

src/main/java/eu/dissco/core/handlemanager/domain/responses/DigitalSpecimenBotanyResponse.java
southeo: I'm aware it's more mutable than desirable. Yes, I instantiate it and then set all the properties. 

Right now, responses are built row by row (see HandleRepository.java for functions). It's a side effect of how the handle record is stored within the database. When we resolve a handle, we have to parse a list of rows to get all the data we need. Each row holds a single property for our Response object.

An alternative I can think of would be to store each element in a specific variable as we loop through the list of HandleAttributes. then initialize the ResponseObject using these explicit variables. Something like

for each attribute:
- type = HANDLES.TYPE
- switch(type)
- case("pidIssuer") pidIssuer = HANDLES.DATA
- case("pidStatus") pidStatus = HANDLES.DATA
- case("issueDate") issueDate = HANDLES.DATA

Response response = new Response(pidIssuer, pidStatus, issueDate, ...)

The more I think about it, the more I think having these specific response objects is a bad idea. I think responses should be jsonnodes instead.

src/main/java/eu/dissco/core/handlemanager/repository/HandleRepository.java
```diff
@@ -1,39 +1,522 @@
...
+    return attributes;
+  }
 
-}
+  // Get List of Pids
+  public List<String> getAllHandles(byte[] pidStatus, int pageNum, int pageSize) {
+    return context
+        .selectDistinct(HANDLES.HANDLE, HANDLES.DATA)
+        .from(HANDLES)
+        .where(HANDLES.DATA.eq(pidStatus)) // Hmm...
```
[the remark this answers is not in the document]
southeo: I'm not 100% sure what you mean, is STATUSTYPE an enum in this example?

src/main/java/eu/dissco/core/handlemanager/repository/HandleRepository.java
```diff
@@ -1,39 +1,522 @@
...
-  // Resolve single handle
-  @Query(value = "select * from handles where handle = ?1", nativeQuery = true)
-  List<Handles> resolveHandle(byte[] handle);
+  // Resolve Pid
+  public List<HandleAttribute> resolveHandle(byte[] handle) {
+    List<Record4<Integer, byte[], byte[], byte[]>> dbRecord = context
+        .select(HANDLES.IDX, HANDLES.HANDLE, HANDLES.TYPE, HANDLES.DATA)
+        .from(HANDLES)
+        .where(HANDLES.HANDLE.eq(handle))
+        .fetch();
```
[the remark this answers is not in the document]
southeo: neat!

pom.xml
```diff
@@ -65,8 +64,23 @@
       <groupId>org.springframework.boot</groupId>
     </dependency>
     <dependency>
-      <artifactId>spring-boot-starter-data-jpa</artifactId>
-      <groupId>org.springframework.boot</groupId>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context</artifactId>
+    </dependency>
+    <dependency>
```
samleeflang, line 70: I think this jdbc driver is conflicting with JOOQ and Postgres, failing the build. Why was it necessary to add it?

samleeflang (approved):

PR #15, opened 2023-02-27
Introduce Trivy Workflow

southeo:
Proposed Trivy Workflow:

1. Scan for all vulnerabilities, list as table
2. Scan for CRITICAL and HIGH risk vulnerabilities. Will break on a vulnerability not in .trivyignore


.github/workflows/build.yaml
```diff
@@ -37,6 +37,26 @@ jobs:
           -Dsonar.projectKey=DiSSCo_${{ github.event.repository.name }}
           -Dsonar.exclusions=**/jooq/**
           -Dsonar.coverage.exclusions=**/properties/**,**/configuration/**,**/domain/**,**/exception/**
+      - name: Build Docker Image
+        run: |
+          docker build -t docker.io/dissco/handle-manager:${{ github.sha }} .
```
samleeflang, line 42: We are building the container now twice in the CI pipeline. Here for Trivy and below for the Push. Could we combine the steps? See if you can merge them so that we build only a single time and push only when we merge to master. We then push two images, one with the :${sha} and the :latest. Also the step could be more generic by using public.ecr.aws/dissco/${{ github.event.repository.name }} as the string for the image.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM eclipse-temurin:17-alpine as builder
 WORKDIR application
-ARG JAR_FILE=target/*.jar
+ARG JAR_FILE=target/*spring-boot.jar
```
samleeflang, line 3: Still strange that this is necessary...

.github/workflows/.trivyignore
```diff
@@ -0,0 +1,5 @@
+# Spring Web
```
samleeflang, line 1: Need to document this well so we don't forget them.

southeo: # Date: Feb-28-2022
# Package: Snakeyaml
# Solution: Upgrading major version of Spring may fix this issue, but will break other dependencies
# Status: User story in Jira

southeo:
Current workflow: 

* Build image
* Scan image (2x)
* Push image to ECR (if appropriate)

The minor issue is that we don't get to make use of a nice and tidy github action (i.e. Build and Push) to push to ECR. We're instead manually pushing under a "run" action. We have to manually tag the image and manually add the labels instead of having Docker Meta do the work for us. 

It might be worthwhile just to rebuild the image if we push to ECR. We only rebuild the image if it's a push to main, so it's not a lot of extra work in the grand scheme.

samleeflang (approved):

PR #17, opened 2023-03-20
Locations attribute is optional in request

southeo:
Two locations are automatically generated based on the handle minted:

`https://sandbox.dissco.tech/api/v1/specimens/ + handle`
`https://sandbox.dissco.tech/ds/" + handle`

When creating a new handle record, users have the option to provide an array of additional locations. 

All weights are 0, except for the API location. (Weights are used if a machine is performing a random selection on which location to redirect to, and we want predictable behaviour). 

The api location has an id of 0, the ui has an id of 1, and additional locations are found at id 2+. 

Example Record: https://hdl.handle.net/20.5000.1025/2Q8-Z4P-AJX?noredirect

[Jira DD-365](https://naturalis.atlassian.net/browse/DD-365?atlOrigin=eyJpIjoiYjVhMzg0NzJmNWIzNDY5ZWFlNThkMDM2MzI2MjI2NjgiLCJwIjoiaiJ9)

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,13 +1,23 @@
-# Date: Feb-28-2022
+# Date: Feb-28-2023
 # Package: Spring Web
+# Issue: "HttpInvokerServiceExporter readRemoteInvocation method untrusted java deserialization"
+# Severity: Critical
 # Notes: Upgrading this package will depreciate other security packages. We need to rethink our AAI
 # Status: User story in Jira (DD-273)
 CVE-2016-1000027
 
-# Date: Feb-28-2022
+# Date: Feb-28-2023
 # Package: Snakeyaml
+# Issue: "Denial of Service due to missing nested depth"
+# Severity: High
 # Solution: This package is managed by Spring. Upgrading major version of Spring may fix this issue, but will break other dependencies
 # Status: User story in Jira (DD-272)
 CVE-2022-25857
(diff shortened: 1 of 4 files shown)


.github/workflows/.trivyignore
```diff
@@ -9,3 +9,5 @@ CVE-2016-1000027
 # Solution: This package is managed by Spring. Upgrading major version of Spring may fix this issue, but will break other dependencies
 # Status: User story in Jira (DD-272)
 CVE-2022-25857
+CVE-2022-1471
```
samleeflang, line 12: What do we need this exception for? Could we include the additional information as we did in the other one, date of adding (and some other things)?

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -698,6 +700,23 @@ public byte[] setLocations(String[] objectLocations) throws PidServiceInternalEr
...
+  }
+
+  private String[] defaultLocations(String handle){
+    String api = "https://sandbox.dissco.tech/api/v1/specimens/" + handle;
+    String ui = "https://sandbox.dissco.tech/ds/" + handle;
+    return new String[]{api, ui};
+  }
+
+
+
```
samleeflang, line 719: Not needed to fix in this PR but the service is getting pretty big

samleeflang (approved):

PR #18, opened 2023-03-31
Resolve external pids

southeo:
Added functionality to resolve external handles, i.e. handles that do not fall under our prefix. This will be useful in, say, resolving the FDO profile to determine which attributes are expected in the record. (see https://hdl.handle.net/21.T11148/d8de0819e144e4096645). 

The web client maps responses to a jsonNode. If the pid does not lead to a json, an UnprocessableEntityException is thrown.

Removed the PidTypeService. This service resolved dissco handles and mapped them to a json. The data model has been flattened, so we don't need this anymore. 

Methods that generate a list of `HandleAttributes` from a request have been moved to their own class, `FdoRecordBuilder`. The service layer was getting very big. 

Added an endpoint to allow end-users to resolve external handles. Not sure if we should keep this endpoint, though. 

[Jira](https://naturalis.atlassian.net/browse/DD-368?atlOrigin=eyJpIjoiZjkxMjNhMWY3ZTk0NDBiYTlhOWM3MTA2NzBhNjU3OWMiLCJwIjoiaiJ9)


src/main/java/eu/dissco/core/handlemanager/component/PidResolverComponent.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.web.reactive.function.client.WebClient;
+
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class PidResolverComponent {
+
+  private final WebClient webClient;
+
+  @Cacheable("pid")
```
samleeflang, line 21: Did you check if it works. Maybe you need to add the `@EnableCaching`

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
```diff
@@ -79,12 +79,28 @@ public ResponseEntity<List<String>> getAllHandlesByPidStatus(
...
-      @PathVariable("suffix") String suffix, HttpServletRequest r) throws PidResolutionException {
+  public ResponseEntity<JsonApiWrapperReadSingle> resolvePid(@PathVariable("prefix") String prefix,
+      @PathVariable("suffix") String suffix, HttpServletRequest r)
+      throws PidResolutionException, UnprocessableEntityException {
+
     String path = SANDBOX_URI + r.getRequestURI();
-    byte[] handle = (prefix + "/" + suffix).getBytes(StandardCharsets.UTF_8);
+    String handle = prefix + "/" + suffix;
+
+    if (prefix.equals("20.5000.1025")){
```
samleeflang, line 89: At some point we need to make the SANDBOX_URI and the internal prefix

src/main/java/eu/dissco/core/handlemanager/component/PidResolverComponent.java
southeo: It works! @EnableCaching is already there. This component is replacing a resolver for internal pids, so we already had caching enabled. :)

src/main/java/eu/dissco/core/handlemanager/component/PidResolverComponent.java
```diff
@@ -0,0 +1,47 @@
...
+    String url = "https://hdl.handle.net/" + pid;
+    log.info("querying the following: {}", url);
+    var responseSpec = webClient.get().uri(url).retrieve()
+        .onStatus(HttpStatus.NOT_FOUND::equals,
+            r -> r.bodyToMono(String.class).map(PidResolutionException::new));
+    var response = responseSpec.bodyToMono(JsonNode.class);
+    JsonNode responseVal;
+
+    try {
+      responseVal = response.toFuture().get();
```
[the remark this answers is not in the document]
southeo: Changed

samleeflang (approved):

PR #20, opened 2023-05-15
Spring boot 3.0

southeo:
[Jira](https://naturalis.atlassian.net/browse/DD-506?atlOrigin=eyJpIjoiYzcwZDFjY2NhYWZjNDZkY2E4ZWY5YmEzNTgwNzkxMDgiLCJwIjoiaiJ9)


pom.xml
```diff
@@ -66,6 +60,12 @@
     <dependency>
       <artifactId>spring-boot-starter-web</artifactId>
       <groupId>org.springframework.boot</groupId>
+      <exclusions>
```
samleeflang, line 63: I think the exclusion is not necessary when you pin the version of snakeYaml

PR #22, opened 2023-06-05
FdoRecordBuilder prepares update handle attribute list

southeo:
When an update request is made, the FdoRecordBuilder creates a list of HandleAttributes to post to the db.


src/main/java/eu/dissco/core/handlemanager/component/FdoRecordBuilder.java
```diff
@@ -437,11 +447,47 @@ public List<HandleAttribute> prepareDigitalSpecimenBotanyRecordAttributes(
...
   }
 
+  public List<HandleAttribute> prepareUpdateAttributes(byte[] handle, JsonNode requestAttributes)
+      throws InvalidRequestException, PidServiceInternalError {
+    requestAttributes = setLocationXmlFromJson(requestAttributes,
+        new String(handle, StandardCharsets.UTF_8));
+    Map<String, String> updateRequestMap = mapper.convertValue(requestAttributes,
+        new TypeReference<Map<String, String>>() {
+        });
+    List<HandleAttribute> attributesToUpdate = new ArrayList<>();
```
samleeflang, line 457: Could use a stream here. Then you don't have to create an empty List and then fill it but just return a list from the stream.

samleeflang (approved):
Looks good, small comment nothing more

PR #26, opened 2023-06-12
Feature/rollback handles

southeo:
Two endpoints to allow the rollback of handles in case of exceptions in the Processing Service

1. Undo Handle Creation -> Need to delete entire handle record
2. Undo Handle Update -> need to revert record to previous version. Post an update to the handle repository, but decrement the version number instead of incrementing it. 

This should be behind a role specific to the processing service.


src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
```diff
@@ -187,6 +188,31 @@ public ResponseEntity<JsonApiWrapperWrite> archiveRecord(@PathVariable("prefix")
     return ResponseEntity.status(HttpStatus.OK).body(service.archiveRecordBatch(List.of(request)));
   }
 
+  @Operation(summary = "rollback handle creation")
+  @DeleteMapping(value="/rollback")
+  public ResponseEntity<Void> rollbackHandleCreation(@RequestBody RollbackRequest request)
+      throws InvalidRequestException {
+
+    var ids = request.data().stream().map(d -> d.get(NODE_ID)).toList();
```
samleeflang, line 196: Might want to log who is requesting what. Something like "Request for rollback creation has been received for PID: pid from user: X". Noticed we are missing those at other calls as well. Might be something for now or a task to add later.

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
```diff
@@ -187,6 +188,31 @@ public ResponseEntity<JsonApiWrapperWrite> archiveRecord(@PathVariable("prefix")
...
+    if (ids.contains(null)){
+      throw new InvalidRequestException("Missing Handles (\"id\") in request");
+    }
+    var handles = ids.stream().map(JsonNode::asText).toList();
```
samleeflang, line 200: Why the full JsonNode, can't the RollbackRequest not be just a list of Strings? Or is there more info in the JsonNode, maybe then we should do a JsonSchema validation?

src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
```diff
@@ -36,33 +36,32 @@
...
-import eu.dissco.core.handlemanager.domain.requests.attributes.PidStatus;
 import eu.dissco.core.handlemanager.domain.requests.validation.JsonSchemaValidator;
 import eu.dissco.core.handlemanager.exceptions.InvalidRequestException;
 import eu.dissco.core.handlemanager.exceptions.PidResolutionException;
 import eu.dissco.core.handlemanager.service.HandleService;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import lombok.extern.slf4j.Slf4j;
```
samleeflang, line 57: Is this import necessary?

samleeflang (commented):
Looks good!

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
southeo: We could use the JsonSchema validation, but this is simple enough that I think we don't need it. 

JsonApi wants us to have:

```
{
   "data": [
      {"id":"20.500/123"},
      {"id":"20.500/555}
   ]
}
```
So I extract all the "id" nodes from the data node. `ids` is a List < JsonNode >  of `[ {"20.500/123"}, {"20.500/555"}]`

Then we stream() that list and get a List < String > . 

It's possible the request contains additional information, but we just ignore it. 

It WOULD be a lot easier to just accept a List < String > in the request body, but when has JsonSchema ever been easy :)

src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
southeo: noo

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
southeo: Hmm, we could potentially rollback 500 handles at once (or whatever the specimen processor max is). Is that going to clutter the log too much? 

We also don't capture user information right now.

samleeflang: Nah I think it will be ok and hopefully we won't rollback too often and when we do it might be good to have some idea which ones we rolled back. And 500 lines is not much. To capture user info just add `Authentication authentication` as a function param, just as the backend does.

src/main/java/eu/dissco/core/handlemanager/controller/HandleController.java
samleeflang: Yes you are right... just looks like such as hassle. All the JsonNode are not really how Java strong-typing should work but anyway it does provide a lot of flexibility.

samleeflang (approved):

PR #27, opened 2023-06-13
Feature/more object types

southeo:
New objects can be created through the Handle API. Note that the FDO Profiles for these objects are still in development.

New Object Types added:
- annotation
- organisation
- mapping
- sourceSystem

Additionally, mediaObject has been updated to fit with the most up to-date RFC. 

[RFC Documents](https://drive.google.com/drive/folders/1FkxarExjVI4MpL_TdBHXxu4G6IhOagPx?usp=sharing)
[Jira](https://naturalis.atlassian.net/browse/DD-511?atlOrigin=eyJpIjoiZWJkY2U2MTJkZDBkNDNiZmFhMDA3Y2JjOWI3MGI2ZTIiLCJwIjoiaiJ9)


src/main/java/eu/dissco/core/handlemanager/component/FdoRecordBuilder.java
```diff
@@ -516,18 +642,40 @@ private byte[] setLocations(String[] userLocations, String handle) throws PidSer
...
+      case DIGITAL_SPECIMEN -> {
+        String api = appProperties.getApiUrl() + "/specimens/" + handle;
+        String ui = appProperties.getUiUrl() + "/ds/" + handle;
+        return new String[]{api, ui};
+      }
+      case MAPPING -> {
+        return new String[]{appProperties.getApiUrl() + "/mapping/" + handle};
+      }
+      case SOURCE_SYSTEM -> {
+        return new String[]{appProperties.getApiUrl() + "/source-system/" + handle};
```
samleeflang, line 664: We have to figure out where to place the orchestration endpoints. Currently they are not at the same API as the other resources. However, we have been running into CORS and it might be better to have them on the same place as the rest.

src/main/java/eu/dissco/core/handlemanager/domain/requests/vocabulary/ObjectType.java
```diff
@@ -8,6 +8,11 @@ public enum ObjectType {
   @JsonProperty("digitalSpecimen") DIGITAL_SPECIMEN("digitalSpecimen"),
   @JsonProperty("digitalSpecimenBotany") DIGITAL_SPECIMEN_BOTANY("digitalSpecimenBotany"),
   @JsonProperty("mediaObject") MEDIA_OBJECT("mediaObject"),
+  @JsonProperty("annotation") ANNOTATION("annotation"),
+  @JsonProperty("sourceSystem") SOURCE_SYSTEM("sourceSystem"),
+  @JsonProperty("mapping") MAPPING("mapping"),
+  @JsonProperty("translator") TRANSLATOR("translator"),
```
samleeflang, line 14: Is translator an object?

samleeflang (commented):
Looks nice! Smart to use the objectType and then only generate the additional attributes needed

src/main/java/eu/dissco/core/handlemanager/domain/requests/vocabulary/ObjectType.java
southeo: No it's not. removed it.

src/main/java/eu/dissco/core/handlemanager/component/FdoRecordBuilder.java
southeo: changed to orchestration.dissco.tech/api/v1

samleeflang (approved):

PR #28, opened 2023-06-15
fix caching

southeo:
A method's response is not cached if it is called from within the same class. This has been fixed.


Dockerfile
```diff
@@ -1,4 +1,7 @@
-FROM eclipse-temurin:17-alpine as builder
+FROM eclipse-temurin:17-alpine AS builder
+RUN apk update
+RUN apk add "openssl=3.1.1-r1"
```
samleeflang, line 3: Make a comment or put in some reminder that at some point we should remove this.

samleeflang (approved):

PR #33, opened 2023-07-05
Feature/enumerate pid profile

southeo:
Change FDO profile attributes to an enum to allow for better index coupling.


samleeflang (approved):
Nice, looks good!
I would remove the comments and put this in Jira tickets.

PR #41, opened 2023-07-25
Feature/remove botany specimen

southeo:
Remove botany specimen requests. Different specimen types will have the same fdo profiles.


samleeflang (approved):
:balloon:

PR #46, opened 2023-08-16
Feature/smaller upsert responses

southeo:
Upsert endpoint (used by specimen processor) returns very large objects. We only need to return the newly minted handle and the primaryPhysicalSpecimenObjectId. 

New Response looks like: 

`
{
    "data": [

        {

            "id": "20.5000.1025/FSL-ME2-9ZJ",

            "type": "digitalSpecimen",

            "attributes": {

                "primarySpecimenObjectId": "localId1"

            },

            "links": {

                "self": "https://hdl.handle.net/20.5000.1025/FSL-ME2-9ZJ"

            }

        },

        {

            "id": "20.5000.1025/0AL-NRE-7Y7",

            "type": "digitalSpecimen",

            "attributes": {

                "primarySpecimenObjectId": "localId2"

            },

            "links": {

                "self": "https://hdl.handle.net/20.5000.1025/0AL-NRE-7Y7"

            }

        }

    ]
}

`

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/repository/HandleRepository.java
+++ b/src/main/java/eu/dissco/core/handlemanager/repository/HandleRepository.java
@@ -71,6 +71,15 @@ public class HandleRepository {
         .fetch(this::mapToAttribute);
   }
 
+  public List<HandleAttribute> getPrimarySpecimenObjectId(List<byte[]> handles){
+    return context
+        .select(HANDLES.IDX, HANDLES.HANDLE, HANDLES.TYPE, HANDLES.DATA)
+        .from(HANDLES)
+        .where(HANDLES.HANDLE.in(handles))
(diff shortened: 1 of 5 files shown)


src/main/java/eu/dissco/core/handlemanager/repository/HandleRepository.java
```diff
@@ -71,6 +71,15 @@ public List<HandleAttribute> searchByNormalisedPhysicalIdentifier(List<byte[]> n
         .fetch(this::mapToAttribute);
   }
 
+  public List<HandleAttribute> getPrimarySpecimenObjectId(List<byte[]> handles){
```
samleeflang, line 74: Do we really need a new db call? Can't we get all the info we need from the stuff we already have in mem?

southeo: You're right. This is done elsewhere as well to make formatting responses easier. I'll make a new branch to fix that.

samleeflang (approved):

PR #47, opened 2023-08-21
Qnumber support

southeo:
Resolve qnumbers for specimenHosts

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
@@ -105,6 +105,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 import org.w3c.dom.Document;
+import org.webjars.NotFoundException;
 
 @RequiredArgsConstructor
 @Service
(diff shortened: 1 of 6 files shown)


src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -585,18 +587,25 @@ private List<HandleAttribute> setSpecimenHostName(DigitalSpecimenRequest request
               SPECIMEN_HOST_NAME.get(),
               specimenHostName.getBytes(StandardCharsets.UTF_8)));
     } else {
+      String specimenHostNameResolved;
       try {
-        var specimenHostNameResolved = pidResolver.getObjectName(getRor(request.getSpecimenHost()));
+        String specimenHostId = request.getSpecimenHost();
+        if(specimenHostId.contains(ROR_DOMAIN)){
+          specimenHostNameResolved = pidResolver.getObjectName(getRor(specimenHostId));
+        } else{
```
samleeflang, line 595: Maybe also put an else if here for the WikiData and throw an error when anything else.

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -585,18 +587,25 @@ private List<HandleAttribute> setSpecimenHostName(DigitalSpecimenRequest request
...
         fdoRecord.add(
             new HandleAttribute(SPECIMEN_HOST_NAME.index(), handle,
                 SPECIMEN_HOST_NAME.get(),
                 specimenHostNameResolved.getBytes(StandardCharsets.UTF_8)));
-
-      } catch (Exception e) {
+      } catch (NotFoundException | UnprocessableEntityException | InvalidRequestException |
+               PidResolutionException e ) {
         log.error(
             "SpecimenHostId is not a resolvable ROR and no SpecimenHostName is provided in the request. SpecimenHostName field left blank. More information: "
```
samleeflang, line 605: Do we need to mention WikiData in the log message?

samleeflang (commented):
Nice, only a small comment

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
southeo: Added log message to the webclient class (pid resolver).

samleeflang (approved):

PR #48, opened 2023-08-23
remove second db call for formatting

southeo:
Remove second call for formatting
Update and archive responses are smaller; only changed attributes are returned.

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
@@ -7,6 +7,7 @@ import static eu.dissco.core.handlemanager.domain.JsonApiFields.NODE_DATA;
 import static eu.dissco.core.handlemanager.domain.JsonApiFields.NODE_ID;
 import static eu.dissco.core.handlemanager.domain.JsonApiFields.NODE_TYPE;
 import static eu.dissco.core.handlemanager.domain.requests.vocabulary.ObjectType.DIGITAL_SPECIMEN;
+import static eu.dissco.core.handlemanager.domain.requests.vocabulary.ObjectType.TOMBSTONE;
 import static eu.dissco.core.handlemanager.service.ServiceUtils.setUniquePhysicalIdentifierId;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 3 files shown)


src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -74,45 +78,41 @@ public JsonApiWrapperReadSingle resolveSingleRecord(byte[] handle, String path)
...
     }
     return new JsonApiWrapperRead(new JsonApiLinks(path), dataList);
   }
 
+  private void verifyHandleResolution(List<byte[]> handles, List<HandleAttribute> dbRecords) throws PidResolutionException {
+    var resolvedHandles = dbRecords.stream()
+        .map(HandleAttribute::handle)
+        .map(handle -> new String(handle, StandardCharsets.UTF_8))
+        .collect(Collectors.toSet());
+    if (handles.size()==resolvedHandles.size()){
```
samleeflang, line 95: format looks off. Check if this is something that might be interesting. I might start using it myself as well.

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -74,45 +78,41 @@ public JsonApiWrapperReadSingle resolveSingleRecord(byte[] handle, String path)
...
+      return;
+    }
+    var handlesString = handles.stream().map(handle -> new String(handle, StandardCharsets.UTF_8)).collect(
```
samleeflang, line 98: I like the formatting of the other stream, could we format this the same way?

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -558,33 +571,26 @@ public JsonApiWrapperWrite archiveRecordBatch(List<JsonNode> requests)
       throws InvalidRequestException, PidResolutionException, PidServiceInternalError, UnprocessableEntityException {
     var recordTimestamp = Instant.now().getEpochSecond();
     List<byte[]> handles = new ArrayList<>();
-    List<HandleAttribute> archiveAttributes = new ArrayList<>();
+    List<HandleAttribute> archiveAttributesFlat = new ArrayList<>();
+    List<List<HandleAttribute>> archiveAttributes = new ArrayList<>();
```
samleeflang, line 575: Could use `var` keyword

samleeflang (commented):
Just formatting nitpicking

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
samleeflang: https://www.jetbrains.com/help/idea/reformat-and-rearrange-code.html#reformat-on-save

PR #49, opened 2023-08-23
Rollback endpoint added

southeo:
Added endpoint to remove handles if you do not know the Handle, only the primary physical specimen object id. 

This is used in the event of a dropped connection/memory overflow. e.g. Processing service requests creation of handles, handle api creates handles, processing service can't process response from handle api. In this case, the handles are created but we don't know what the handles are, so we delete by the primary specimen object Id.

### DIFF
--- a/src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
+++ b/src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
@@ -81,7 +81,6 @@ class HandleControllerTest {
   @Mock
   private JsonSchemaValidator schemaValidator;
 
-
   private HandleController controller;
 
   private final String SANDBOX_URI = "https://sandbox.dissco.tech";
@@ -116,7 +115,7 @@ class HandleControllerTest {
   }
 
   @Test
-  void testResolvePidBadPrefix(){
+  void testResolvePidBadPrefix() {
     assertThrows(PidResolutionException.class, () ->
         controller.resolvePid(SUFFIX, SUFFIX, new MockHttpServletRequest()));
   }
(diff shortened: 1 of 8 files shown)


src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
```diff
@@ -127,11 +126,13 @@ void testSearchByPhysicalId() throws Exception {
...
-        SPECIMEN_HOST_TESTVAL)).willReturn(responseExpected);
+    given(
+        service.searchByPhysicalSpecimenId(PRIMARY_SPECIMEN_OBJECT_ID_TESTVAL, PhysicalIdType.CETAF,
+            SPECIMEN_HOST_TESTVAL)).willReturn(responseExpected);
 
     // When
-    var responseReceived = controller.searchByPhysicalSpecimenId(PRIMARY_SPECIMEN_OBJECT_ID_TESTVAL, PhysicalIdType.CETAF,
+    var responseReceived = controller.searchByPhysicalSpecimenId(PRIMARY_SPECIMEN_OBJECT_ID_TESTVAL,
+        PhysicalIdType.CETAF,
         SPECIMEN_HOST_TESTVAL);
```
samleeflang, line 136: I think this can move to the previous line

src/test/java/eu/dissco/core/handlemanager/controller/HandleControllerTest.java
```diff
@@ -548,28 +555,42 @@ void testRollbackHandles() throws Exception{
...
 
+  @Test
+  void testRollbackHandlesByPhysId() {
+    // Given
+    given(authentication.getName()).willReturn("name");
+    var physIds = List.of("a", "b");
+
+    // When
+    controller.rollbackHandlePhysId(physIds, authentication);
+
```
samleeflang, line 573: // Then

samleeflang (approved):
Two really small comments. Approved so when it is changed you can just merge

PR #51, opened 2023-08-28
Remove admin handle from response

southeo:
Remove HS admin handle from the JSON response.

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
@@ -1,5 +1,6 @@
 package eu.dissco.core.handlemanager.service;
 
+import static eu.dissco.core.handlemanager.domain.FdoProfile.HS_ADMIN;
 import static eu.dissco.core.handlemanager.domain.FdoProfile.PID;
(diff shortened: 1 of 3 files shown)


src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -108,13 +108,24 @@ private void verifyHandleResolution(List<byte[]> handles, List<HandleAttribute>
...
     var handleMap = mapRecords(dbRecord);
+    removeHsAdmin(handleMap);
     List<JsonNode> rootNodeList = new ArrayList<>();
     for (var handleRecord : handleMap.entrySet()) {
       rootNodeList.add(jsonFormatSingleRecord(handleRecord.getValue()));
     }
     return rootNodeList;
   }
 
+  private Map<String, List<HandleAttribute>> removeHsAdmin(
```
samleeflang, line 119: Not sure if we need an extra loop here. Could we not include this in the jsonFormatSingleRecord? That would save us an additional loop over all records.

southeo: Good call, made the switch

samleeflang (approved):

PR #53, opened 2023-08-28
Align with up new fdo profile

southeo:
- physicalIdType can now be local, global, or resolvable. removed cetaf/combined options. 
- Normalised Primary Physical Object Id is now generated using source system id (if physical id is not global/resolvable)
- added topicCategory to fdo profile 
- otherSpecimenIds is now an object, not just a list of strings
- Topic category/domain/discipline/origin, BaseTypeOfSpecimen, MaterialOrDigitalEntity, MaterialSampleType are enforced vocabularies
- StructuralType is a vocabulary. All objects within dissco currently are of StructuralType "digital", but there may be other object types in the future which are physical, abstractions, or performances. 

A given topicCategory can only be added to specific topicDisciplines (e.g. topicCategory.amphibian can only be in topicDiscipline.Zoology)

A given materialSampleType can only be added to specific topicDisciplines/topicDomain/topicOrigins. e.g. materialSampleType.Whole_Organism can be added to topicDiscipline.zoology or topicDiscipline.botany. For materialSampleType.artefact, there is no corresponding discipline, but there is a corresponding domain (topicCategory.Cultural_Artefact). 

Also removed some objects that aren't used any more


src/main/java/eu/dissco/core/handlemanager/domain/requests/vocabulary/MaterialSampleType.java
```diff
@@ -0,0 +1,34 @@
+package eu.dissco.core.handlemanager.domain.requests.vocabulary;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public enum MaterialSampleType {
```
samleeflang, line 5: Does make me wonder how we are going to determine this but that is something for later

src/main/java/eu/dissco/core/handlemanager/domain/requests/vocabulary/TopicDiscipline.java
```diff
@@ -0,0 +1,146 @@
...
+import static eu.dissco.core.handlemanager.domain.requests.vocabulary.TopicCategory.SPONGES;
+import static eu.dissco.core.handlemanager.domain.requests.vocabulary.TopicCategory.VERT_FOSSILS;
+import static eu.dissco.core.handlemanager.domain.requests.vocabulary.TopicCategory.VIRUSES;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Collections;
+import java.util.Set;
+import lombok.Getter;
+
+public enum TopicDiscipline {
```
samleeflang, line 65: Hmm this differs from the ones we use within DiSSCo translator/frontend. Might need to have a look

samleeflang (commented):
Looking great, very happy about the enums. I do wondering if the normailised id should be generate by the handle as we will also this in the translator service. That might start to conflict. I think the normalised id the Handle API should just take over/require. 
Couple things might require us to take a look when we updated the translator but nothing major I think

samleeflang (approved):

PR #54, opened 2023-09-01
Feature/media fdo profile

southeo:
- Added some vocabularies (mediaFormat, linkedDigitalObjectType, primaryMediaObjectIdType, primaryMediaObjectType)
- Changed HandleAttribute to a POJO from a record. allows for a much cleaner constructor, but it means there are a lot of lines touched by this change 
- [RFC](https://docs.google.com/document/d/17wsPgTL83bov9LAX5GaY_kpKQAn3AFOP7TkNVKXh0Z0/edit)


src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
```diff
@@ -51,12 +51,26 @@ public enum FdoProfile {
   WAS_DERIVED_FROM_ENTITY("wasDerivedFromEntity", 218),
 
   // Media
-  MEDIA_URL("mediaUrl", 403),
-  MEDIA_OBJECT_TYPE("mediaObjectType", 405),
-  MEDIA_HOST("mediaHost", 406),
-  SUBJECT_LOCAL_ID("subjectLocalIdentifier", 407),
-  SUBJECT_PID("subjectPid", 408),
-
+  MEDIA_HOST("mediaHost", 400),
```
samleeflang, line 54: These come from the approved RFC right?

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -284,24 +276,79 @@ public List<HandleAttribute> prepareMediaObjectAttributes(MediaObjectRequest req
...
+      fdoRecord.add(new HandleAttribute(MEDIA_FORMAT, handle, request.getMediaFormat().toString()));
+    }
+    fdoRecord.add(new HandleAttribute(IS_DERIVED_FROM_SPECIMEN, handle,
+        request.getIsDerivedFromSpecimen().toString()));
+    fdoRecord.add(
+        new HandleAttribute(LINKED_DO_PID, handle, request.getLinkedDigitalObjectPid()));
+    fdoRecord.add(
+        new HandleAttribute(LINKED_DO_TYPE, handle, request.getLinkedDigitalObjectType()
+            .toString()));
+    if (request.getLinkedAttribute() != null) {
```
samleeflang, line 295: So we only add it if it is not null, otherwise the whole key won't be visible?

samleeflang (commented):
What happens when we deploy this? Would the current media processor still work or would these changes break the flow?

src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
southeo: Yes, all from the RFC

southeo:
> What happens when we deploy this? Would the current media processor still work or would these changes break the flow?

Media processor would break, need to make minor changes. That's in the works still.

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
southeo: Not sure what you mean by "the whole key." `request.getXYZ()` gets the value of the attribute, not the key. the key is handle + index (index is defined in the FdoProfile enum `LINKED_DO_TYPE`). 

If we didn't do the null check, we would wind up with blank values in the handle record. Since these fields are optional, I thought it would be better to omit the field than leave them blank. 

Does that answer?

samleeflang:
> > What happens when we deploy this? Would the current media processor still work or would these changes break the flow?
> 
> Media processor would break, need to make minor changes. That's in the works still.

Is this still true, maybe wait with merge until everything keeps working after this change

southeo:
> > > What happens when we deploy this? Would the current media processor still work or would these changes break the flow?
> > 
> > 
> > Media processor would break, need to make minor changes. That's in the works still.
> 
> Is this still true, maybe wait with merge until everything keeps working after this change

This branch will work with the PR on the media processor (https://github.com/DiSSCo/dissco-core-digital-media-object-processor/pull/21). They need to be merged at the same time.

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -327,9 +320,11 @@ public List<HandleAttribute> prepareMediaObjectAttributes(MediaObjectRequest req
...
           new HandleAttribute(DERIVED_FROM_ENTITY, handle, request.getDerivedFromEntity()));
     }
-    fdoRecord.add(new HandleAttribute(LICENSE_NAME, handle, request.getLicenseName()));
-    if (request.getLicenseUrl() != null) {
-      fdoRecord.add(new HandleAttribute(LICENSE_URL, handle, request.getLicenseUrl()));
+    if (request.getLicenseName() != null) {
+      fdoRecord.add(new HandleAttribute(LICENSE_NAME, handle, request.getLicenseName()));
+    }
+    if (request.getLicense() != null) {
+      fdoRecord.add(new HandleAttribute(LICENSE_URL, handle, request.getLicense()));
```
samleeflang, line 327: Shouldn't this also be renamed to `LICENSE`?

samleeflang (approved):
One small comment but not blocking.

PR #58, opened 2023-09-05
Shrink create response

southeo:
Smaller responses for digital specimens and media objects, on create endpoint only

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
@@ -4,6 +4,7 @@ import static eu.dissco.core.handlemanager.domain.FdoProfile.HS_ADMIN;
 import static eu.dissco.core.handlemanager.domain.FdoProfile.MEDIA_URL;
 import static eu.dissco.core.handlemanager.domain.FdoProfile.PID;
 import static eu.dissco.core.handlemanager.domain.FdoProfile.PRIMARY_SPECIMEN_OBJECT_ID;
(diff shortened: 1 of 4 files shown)


src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
```diff
@@ -162,10 +163,47 @@ private JsonNode jsonFormatSingleRecord(List<HandleAttribute> dbRecord) {
...
+                DIGITAL_SPECIMEN));
+      } else {
+        var rootNode = jsonFormatSingleRecord(handleRecord.getValue());
+        dataLinksList.add(wrapData(rootNode, type.toString()));
+      }
+    }
+    return dataLinksList;
+  }
+
+  private List<JsonApiDataLinks> formatSimpleResponse(List<HandleAttribute> records,
```
samleeflang, line 190: Can this return more than one attribute? It looks like it get's asked for one returnAttribute (single) so it should always return only the simpleResponse for that attribute. But I am probably misinterpreting this.

samleeflang (commented):
One question

src/main/java/eu/dissco/core/handlemanager/service/HandleService.java
southeo: For digital specimens, it returns the primary specimen object id (aka the physical id). for media objects, it returns the media url. 

Note: the processing service makes handles using the upsert endpoint, not the create endpoint. but it is still possible to use the create endpoint to make a handle for a digital specimen (at the moment anyway)

samleeflang (approved):

PR #60, opened 2023-09-12
Feature/enum error logging

southeo:
Identify the problem enum value (the value that was rejected)

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
+++ b/src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
@@ -330,7 +330,7 @@ public class JsonSchemaValidator {
   }
 
   private String setErrorMessage(Set<ValidationMessage> validationErrors, String type) {
-    return setErrorMessage(validationErrors, type);
+    return setErrorMessage(validationErrors, type, null);
   }
 
   private String setErrorMessage(Set<ValidationMessage> validationErrors, String type,
--- a/src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
+++ b/src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
@@ -27,17 +27,16 @@ import eu.dissco.core.handlemanager.domain.requests.objects.HandleRecordRequest;
 import eu.dissco.core.handlemanager.domain.requests.objects.MappingRequest;
 import eu.dissco.core.handlemanager.domain.requests.objects.MasRequest;
 import eu.dissco.core.handlemanager.domain.requests.objects.MediaObjectRequest;
-import eu.dissco.core.handlemanager.domain.requests.vocabulary.ObjectType;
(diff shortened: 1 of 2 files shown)


src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
```diff
@@ -257,7 +255,7 @@ private void setJsonSchemas() {
   public void validatePostRequest(JsonNode requestRoot) throws InvalidRequestException {
     var validationErrors = postReqSchema.validate(requestRoot);
     if (!validationErrors.isEmpty()) {
-      throw new InvalidRequestException(setErrorMessage(validationErrors, "POST"));
+      throw new InvalidRequestException(setErrorMessage(validationErrors, "POST", null));
```
samleeflang, line 258: Why do these need a `null` field?

src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
```diff
@@ -362,4 +366,13 @@ private String setErrorMessage(Set<ValidationMessage> validationErrors, String t
     return message;
   }
 
+  private String getProblemEnumValue(JsonNode request, String path) {
```
samleeflang, line 369: Would be good to have a test for this

samleeflang (commented):
Small comment, it looks like test coverage is not working is that correct? Or just this PR?

src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
southeo: Just a quick way to reuse code. 

We only eed to add the attributes node to the Error Message function so we can find problem enums values, in the event of an enum error. However, the root request validation functions just make sure the request is in the right format: 
```
{
  "data": {
    "type":"digitalSpecimen", 
    "attributes" : {}
  }
}
```

There will never be an enum error for these top-level checks - it will either be a field missing or an unrecognized property error. So we can pass null to the function safely.

src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
southeo: added a check to existing test

src/main/java/eu/dissco/core/handlemanager/domain/requests/validation/JsonSchemaValidator.java
southeo: To be 100% sure I added an NPE check later but I don't think it's possible to hit.

samleeflang: You could overload the method that might look nicer in the code base? This looks a bit strange, providing `null` as param most of the times.

samleeflang (approved):
Yes much better

PR #62, opened 2023-09-29
add doi profile

southeo:
Added doi profile specifically to set locations for DOIs.

Default Profile behaviour: build 10320/loc attribute urls based on minted handle, and potentially user-inputted locations
DOI Profile behaviour: only use user inputted locations. otherwise it would point to something like https://sandbox.dissco.tech/ds.10.555/qwe-8sq-2, which doesn't exist. 

More features should be added to DOI profile, that's a next week thing

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
@@ -101,6 +101,7 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -733,7 +736,11 @@ public byte[] setLocations(String[] userLocations, String handle, ObjectType typ
   }
 
   private String[] concatLocations(String[] userLocations, String handle, ObjectType type) {
-    ArrayList<String> objectLocations = new ArrayList<>(List.of(defaultLocations(handle, type)));
+    ArrayList<String> objectLocations = new ArrayList<>();
+
+    if (!Arrays.asList(env.getActiveProfiles()).contains("doi")) {
```
samleeflang, line 741: There is this method on the Environment class to check if the profile is active. [matchesProfiles](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html#matchesProfiles(java.lang.String...))

samleeflang, line 741: Make the profile a static variable in a Profile class at the project root so it is clear which profiles there are

samleeflang (approved):

PR #63, opened 2023-09-29
Move profile properties to its own class

southeo:
Some strings will need to change based on if we're minting DOIs or Handles. This class can be useful in keeping them all together.


TomDijkema (approved):
Goedgekeurd

PR #65, opened 2023-10-03
Feature/split doi

southeo:
Partition service into Handle Service and DOI Service. Handle Service functions the same. DOI has Type restrictions (can only create DOIs for Digital Specimens or Media Objects. 

DOI and Handle Services both inherit from the same abstract PidService class. Most of what was in HandleService was moved to PidService. I primise it's not as big a PR as it seems. 

Renamed controller, repository, nameGeneratorService to PidController, PidRepository, PidNameGeneratorService. 

Remarks: 
- We have an Upsert function for Digital specimens but not Media Objects. It's probably not necessary (using the create endpoint is sufficient in the media processor), but it might be nice symmetry. 
- The updates do at least one database call to check if the pid record is writable (i.e it exists and the pidStatus isn't "archived"). [Jira Ticket](https://naturalis.atlassian.net/jira/software/c/projects/DD/issues/DD-724?filter=myopenissues) to see if we can get more out of this database call.
- Lots of internal variable names reference handles. That's fine. I changed logging/exception messages to "identifiers" where possible to provide general responses to the client (e.g. so we don't see the message "minted 100 handles" when we're minting DOIs).


src/main/java/eu/dissco/core/handlemanager/service/DoiService.java
```diff
@@ -0,0 +1,118 @@
...
+    log.info("Persisting new DOIs to db");
+    pidRepository.postAttributesToDb(recordTimestamp, handleAttributes);
+
+    return new JsonApiWrapperWrite(formatCreateRecords(handleAttributes, recordTypes));
+  }
+
+  @Override
+  public JsonApiWrapperWrite updateRecords(List<JsonNode> requests, boolean incrementVersion)
+      throws InvalidRequestException, PidResolutionException, PidServiceInternalError, UnprocessableEntityException {
+    var types = requests.stream()
```
samleeflang, line 104: Do we need to indicate that we might have dropped any record with a different type? Or maybe at least log that we removed X amount of records with a different type. Then we at least know that it is happening

src/main/java/eu/dissco/core/handlemanager/service/PidService.java
```diff
@@ -0,0 +1,527 @@
...
+    if (!registeredSpecimens.isEmpty()) {
+      var registeredHandles = listHandleNamesReturnedFromQuery(registeredSpecimens);
+      throw new PidCreationException(
+          "Unable to create PID records. Some requested records are already registered. Verify the following digital specimens:"
+              + registeredHandles);
+    }
+  }
+
+  // Upsert
+
```
samleeflang, line 279: Remove white line

src/main/java/eu/dissco/core/handlemanager/service/PidService.java
```diff
@@ -0,0 +1,527 @@
...
+      ArrayList<HandleAttribute> upsertAttributeSingleSpecimen = new ArrayList<>(fdoRecordService
+          .prepareUpdateAttributes(upsertRequest.handle().getBytes(StandardCharsets.UTF_8),
+              mapper.valueToTree(upsertRequest.request()), DIGITAL_SPECIMEN));
+      upsertAttributes.add(upsertAttributeSingleSpecimen);
+    }
+    return upsertAttributes;
+  }
+
+  // Update
+
```
samleeflang, line 411: Remove whiteline or the comment all together, it is clear from the method name that this is the update

src/test/java/eu/dissco/core/handlemanager/service/HandleServiceTest.java
```diff
@@ -77,65 +77,38 @@
...
-  @Mock
-  private HandleGeneratorService hgService;
-  @Mock
-  private ProfileProperties profileProperties;
-  private HandleService service;
-  private List<byte[]> handles;
-  private MockedStatic<Instant> mockedStatic;
-  private MockedStatic<Clock> mockedClock;
+@ActiveProfiles(profiles = Profiles.HANDLE)
+class HandleServiceTest extends PidServiceTest {
```
samleeflang, line 92: Does it need to extend the PidServiceTest? You can just instantiate the HandleService and run all tests?

src/test/java/eu/dissco/core/handlemanager/service/PidServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.MockedStatic;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+abstract class PidServiceTest {
```
samleeflang, line 25: This doesn't really test anything, might be better to keep the handle and doi service as separate test classes.

samleeflang (commented):
Couple comment but is looking good

src/main/java/eu/dissco/core/handlemanager/service/DoiService.java
southeo: What this does is collect any Type that isn't DigitalSpecimen or MediaObject. If we have any incompatible types in the request at all, we throw an error (with the problematic types).

I suppose we could filter out the DS/MO objects, but I think if a client is sending a request with mixed types like this, they're probably using the wrong endpoint.

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -18,15 +18,15 @@
 @Service
 @RequiredArgsConstructor
 @Slf4j
-public class HandleGeneratorService {
+public class PidNameGeneratorService {
```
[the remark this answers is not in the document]
southeo: Eh I think "PidName" is more clear. "PidGeneratorService" could be confused for "PidService". the PidService DOES generate pids if you think of the PID as an identifier + the record

src/test/java/eu/dissco/core/handlemanager/service/PidServiceTest.java
southeo: removed

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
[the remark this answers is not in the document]
samleeflang: With name we often refer to a human readable name instead of an id. InstituteId and InstituteName for example that is why I didn't fit to me. But PidIDGeneratorService is also a bit strange. Let's keep this for now

samleeflang (approved):

PR #67, opened 2023-10-04
Update Digital Specimen FDO Profile Again

southeo:
Update Digital Specimen FDO as discussed

Data Model
- LivingOrPreserved: vocab changed to "Living", "
- "livingOrPreserved" attribute name (json parsing) changed so that it's consistent with the class
- referentType is no longer an optional request parameter. It will instead reflect the "type" parameter, so it will be "digitalSpecimen" for DS, "mediaObject" for MO, "organisation" for organisations. 
- normalisedPrimaryObjectId is now a required field. Source system is no longer a request parameter
- added collectionIdentifier as optional parameter

Other 
- Removed redundant exception
- Returned type of pid record on resolution


src/test/java/eu/dissco/core/handlemanager/domain/DigitalSpecimenRequestTest.java
```diff
@@ -140,10 +80,21 @@ void testMaterialSampleType(MaterialSampleType materialSampleType,
...
+        null,
+        topicOrigin,
+        topicDomain,
+        topicDiscipline,
+        null,
+        null,
         null,
-        null, null, null, materialSampleType, null,
-        null, null, SOURCE_SYSTEM_TESTVAL, null));
+        null,
```
samleeflang, line 93: All those null's aren't a pretty sight. But I guess it was already in there. Next time a builder pattern might be nicer with a .with() instead of an all-args constructor

samleeflang (approved):
Just a small comment on the tests but not necessarily on this PR

src/test/java/eu/dissco/core/handlemanager/domain/DigitalSpecimenRequestTest.java
southeo: Hm might be nice, but it would really just be useful for testing. In the main code, we don't create these objects; they're mapped from request JSON. We don't use this constructor elsewhere

PR #68, opened 2023-10-10
fix data model testing

southeo:
License is optional
Improve FdoRecordService testing


samleeflang (approved):
Think there is some overlap with the previous PR

PR #70, opened 2023-11-21
Feature/binary copy insert

southeo:
Use binary copy insert for quick handle creation

The upsert endpoint will be removed. Once that's removed, all insert operations will be done through `copy`. Updates remain as normal. 

Other minor changes: 
- Adds iterator for list of handles generated instead of popping from the list
- Removes AdminHandleGenerator (we generate the value directly from a string now)


src/main/java/eu/dissco/core/handlemanager/properties/DatabaseProperties.java
```diff
@@ -0,0 +1,22 @@
...
+
+import jakarta.validation.constraints.NotBlank;
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@Validated
+@ConfigurationProperties("spring.datasource")
+public class DatabaseProperties {
```
samleeflang, line 11: Wasn't it possible to use the default property from spring for this?

src/main/java/eu/dissco/core/handlemanager/repository/BatchInserter.java
```diff
@@ -0,0 +1,38 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class BatchInserter {
+
+  private final CopyManager copyManager;
+
+  public void batchCopy(long recordTimestamp, List<HandleAttribute> handleAttributes)
+      throws PidCreationException {
+    var rows = handleAttributes.stream()
+        .map(attribute -> new HandleFullRow(attribute, recordTimestamp)).toList();
```
samleeflang, line 26: It feels a bit double the HandleAttribute and HandleFullRow. Couldn't we combine these? Otherwise maybe just make a method that takes an attribute and recordTimestamp and parses it into a csvRow.

src/main/java/eu/dissco/core/handlemanager/domain/repsitoryobjects/HandleFullRow.java
```diff
@@ -0,0 +1,41 @@
+package eu.dissco.core.handlemanager.domain.repsitoryobjects;
+
+import java.nio.charset.StandardCharsets;
+
+public class HandleFullRow {
```
samleeflang, line 5: Not sure about this class. It looks this could be just a method as well

src/test/java/eu/dissco/core/handlemanager/repository/BaseRepositoryIT.java
```diff
@@ -3,8 +3,11 @@
 import static org.testcontainers.containers.PostgreSQLContainer.IMAGE;
 
 import com.zaxxer.hikari.HikariDataSource;
+import eu.dissco.core.handlemanager.database.jooq.tables.Handles;
+import eu.dissco.core.handlemanager.domain.repsitoryobjects.HandleAttribute;
 import org.flywaydb.core.Flyway;
 import org.jooq.DSLContext;
+import org.jooq.Record4;
```
samleeflang, line 10: Is this import required?

samleeflang (commented):
Looking good. Did you check how much performance improvement this resulted in? The batch is a bit smaller than I had in my mind, thought we did a bulk of all handles in a single request. So the performance improvement might be smaller than we hoped.

src/main/java/eu/dissco/core/handlemanager/repository/BatchInserter.java
southeo: Changed to a method, I agree that makes more sense

src/test/java/eu/dissco/core/handlemanager/repository/BaseRepositoryIT.java
southeo: Yes, I moved a function into the parent class that uses it. (see line 48)

src/main/java/eu/dissco/core/handlemanager/properties/DatabaseProperties.java
southeo: :exploding_head: that's a lot easier huh

samleeflang (approved):
Very nice improvement, congrats!

PR #71, opened 2023-11-23
Feature/performance improvements

southeo:
- Only 1 Type of object can be received in a POST /batch request. This reduces the number of checks we have to do on each object
- Remove redundant specimen checks (verify specimen doesn't exist in the handle database, verify all specimen ids are unique). these checks are already done in the processing service


src/main/java/eu/dissco/core/handlemanager/service/DoiService.java
```diff
@@ -46,73 +41,43 @@ public DoiService(PidRepository pidRepository,
...
-
-    validateDigitalSpecimens(digitalSpecimenList);
-
-    log.info("Persisting new DOIs to db");
-    var recordTimestamp = Instant.now().getEpochSecond();
-
-    pidRepository.postAttributesToDb(recordTimestamp, handleAttributes);
-
-    return new JsonApiWrapperWrite(formatCreateRecords(handleAttributes, recordTypes));
+    log.info("Persisting new handles to db");
```
samleeflang, line 63: I think the old log message is still correct right? "Persisting new DOIs to db"

samleeflang (commented):
Ok, less familiar with this code so took a bit longer but everything looks ok. Just one comment on a log line. 
I do wonder wether we should get rid of all checks, especially this one:
- "verify specimen doesn't exist in the handle database" concerns me a bit. If we mess up in the specimen processing service we might mess up the DOIs. In addition what if a specimen was registered with the DOI through another service than DiSSCo. It might be worth the performance penalty too keep this check.

samleeflang:
> Re-added the normalised specimen id check

Not seeing the re-adding of the check on already existing. Might have missed it or maybe it wasn't pushed?

southeo:
> > Re-added the normalised specimen id check
> 
> Not seeing the re-adding of the check on already existing. Might have missed it or maybe it wasn't pushed?

Working with sick brain :sweat_smile: just pushed

PR #73, opened 2023-12-12
Upsert specimens

southeo:
Re-adds upsert endpoint. New specimens are Postgres-Copied. Updated specimens are inserted with Jooq. 

We don't verify media objects/annotations haven't been ingested the way we verify specimens. For media/annotation objects, we rely on the respective processing services to verify they haven't been ingested. It might make sense with media objects (since they'll eventually get DOIs), but for annotations, it probably isn't necessary.

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
+++ b/src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
@@ -97,7 +97,7 @@ public enum FdoProfile {
   private final int index;
 
 
-  private FdoProfile(String attribute, int index) {
+  FdoProfile(String attribute, int index) {
     this.attribute = attribute;
     this.index = index;
   }
--- a/src/main/java/eu/dissco/core/handlemanager/service/PidService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/PidService.java
@@ -27,7 +27,9 @@ import eu.dissco.core.handlemanager.domain.jsonapi.JsonApiWrapperReadSingle;
 import eu.dissco.core.handlemanager.domain.jsonapi.JsonApiWrapperWrite;
 import eu.dissco.core.handlemanager.domain.repsitoryobjects.HandleAttribute;
 import eu.dissco.core.handlemanager.domain.requests.objects.DigitalSpecimenRequest;
+import eu.dissco.core.handlemanager.domain.requests.objects.DigitalSpecimenUpdateWrapper;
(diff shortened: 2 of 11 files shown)


samleeflang (commented):
One question about the writable PIDs, might have missed how that part works

src/main/java/eu/dissco/core/handlemanager/service/PidService.java
```diff
@@ -280,36 +282,96 @@ protected ObjectType getObjectType(List<JsonNode> requests) {
...
-          .map(row -> new String(row.getHandle(), StandardCharsets.UTF_8)).toList();
-      log.error("Attempting to register identifiers for existing records");
-      log.debug("Handles already registered: {}", registeredHandles);
-      throw new PidResolutionException(
-          "Unable to create PID records. Some requested records are already registered. Verify the following digital specimens:"
-              + registeredHandles);
+    var registeredPhysicalIdentiferMap = registeredPhysicalIdentifiers.stream()
+        .collect(Collectors.toMap(row -> new String(row.getData(), StandardCharsets.UTF_8),
+            row -> new String(row.getHandle(), StandardCharsets.UTF_8)));
+    var writableHandles = pidRepository.checkHandlesWritable(
```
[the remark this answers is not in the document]
southeo: Checks PID state. If the PID is set to "ARCHIVED", then the specimen will get a new PID instead of updating an archived PID.

src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
```diff
@@ -97,7 +97,7 @@ public enum FdoProfile {
   private final int index;
 
 
-  private FdoProfile(String attribute, int index) {
+  FdoProfile(String attribute, int index) {
```
samleeflang, line 100: Way are these no longer private?

samleeflang (commented):
Good work! Query is a pain but it is a bit faster. Next year start the discussion about the handle server. Just one more question about the no longer private constructions

src/main/java/eu/dissco/core/handlemanager/domain/FdoProfile.java
southeo: "If you do not specify an access modifier the enum constructor it will be implicitly private."

[Source](https://jenkov.com/tutorials/java/enums.html)

Slowly going to remove the private constructors when I touch enum files

samleeflang: Good find :100:

samleeflang (approved):

PR #74, opened 2024-02-19
Exception Recovery

southeo:
Recover if external PID isn't resolved


samleeflang (approved):
Need to discuss this further on thursday

PR #76, opened 2024-03-29
Improve updates

southeo:
Main issue: During updates, we need to check the issue number and increment it by 1. We were doing a separate db read for each object that needed its version to be incremented. now we get all the versions in one read. 

Performance improvement: we go from 12.5 updates / ms to 2.5 updates/ms. about 60% of the time for this is just the update operation to the, it seems to be a pretty expensive operation.

I also removed the "attributes" field in the response. Before, a batch update request would return all attributes in the request (regardless of if they differed from the original record). I don't think that's necessary, and we don't use the response for PID updates in any of the processors. We had an issue with batch responses from the handle api being too big on a POST operation, so I want to be proactive here.

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -2,3 +2,8 @@
 # Issue: Libexpat, parsing large tokens can trigger a denial of service
 # Solution: Update docker image
 CVE-2023-52425
+
+# Date: March 21, 2024
+# Issue: Vulnerability in spring-web
+# Solution: Spring boot needs to update its version of spring
+CVE-2024-22259
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,8 +1,3 @@
-# Date: Feb 19, 2024
-# Issue: Libexpat, parsing large tokens can trigger a denial of service
-# Solution: Update docker image
-CVE-2023-52425
-
 # Date: March 21, 2024
 # Issue: Vulnerability in spring-web
 # Solution: Spring boot needs to update its version of spring
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,4 +1,4 @@
-# Date: Dec 1, 2023
-# Notes: Issue with logback. Spring boot needs to update its version
-# Currently forcing logback version
-# CVE-2023-6378
+# Date: Feb 19, 2024
+# Issue: Libexpat, parsing large tokens can trigger a denial of service
+# Solution: Update docker image
+CVE-2023-52425
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,4 +1,0 @@
-# Date: March 21, 2024
-# Issue: Vulnerability in spring-web
-# Solution: Spring boot needs to update its version of spring
-CVE-2024-22259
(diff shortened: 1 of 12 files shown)


src/test/java/eu/dissco/core/handlemanager/testUtils/TestUtils.java
```diff
@@ -1,22 +1,109 @@
 package eu.dissco.core.handlemanager.testUtils;
 
+import static eu.dissco.core.handlemanager.domain.FdoProfile.ANNOTATION_HASH;
```
southeo, line 3: Sorry did a format here. :upside_down_face:

.github/workflows/.trivyignore
```diff
@@ -1,8 +1,3 @@
-# Date: Feb 19, 2024
-# Issue: Libexpat, parsing large tokens can trigger a denial of service
-# Solution: Update docker image
-CVE-2023-52425
-
 # Date: March 21, 2024
```
samleeflang, line 1: I think this was also resolved in the latest spring boot update

src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
```diff
@@ -151,14 +150,16 @@ public ResponseEntity<JsonApiWrapperWrite> updateRecord(@PathVariable("prefix")
   public ResponseEntity<JsonApiWrapperWrite> updateRecords(@RequestBody List<JsonNode> requests,
       Authentication authentication)
       throws InvalidRequestException, PidResolutionException, UnprocessableEntityException {
+    var start = System.nanoTime();
```
samleeflang, line 153: Should we also have a measurement on the creates? We could also put this under debug?

src/main/java/eu/dissco/core/handlemanager/repository/PidRepository.java
```diff
@@ -35,64 +35,46 @@ public class PidRepository {
 
   // For Handle Name Generation
   public List<byte[]> getHandlesExist(List<byte[]> handles) {
-    return context
-        .selectDistinct(HANDLES.HANDLE)
-        .from(HANDLES)
-        .where(HANDLES.HANDLE.in(handles))
-        .fetch()
-        .getValues(HANDLES.HANDLE, byte[].class);
+    return context.selectDistinct(HANDLES.HANDLE).from(HANDLES).where(HANDLES.HANDLE.in(handles))
```
samleeflang, line 38: I think most changes are formatting changes. It is a bit difficult to see which queries actually changed.

src/main/java/eu/dissco/core/handlemanager/repository/PidRepository.java
```diff
@@ -208,83 +165,98 @@ public void postAndUpdateHandles(long recordTimestamp, List<HandleAttribute> cre
...
+            e -> {
+              incrementByteArray(e.getValue());
+              return e.getValue();
+            }
+        ));
+  }
+
+  private static void incrementByteArray(byte[] version) {
+    for (int i = version.length - 1; i >= 0; i--) {
+      if (version[i] == Byte.MAX_VALUE) { // Check for overflow
```
samleeflang, line 249: Does this happen?

samleeflang (commented):
Great that you were able to improve some things! Just had a few questions

src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
southeo: Whoops this was meant to be deleted. Don't think this is necessary on updates or creates tbh

PR #77, opened 2024-04-05
Evict cache every 12 hours

southeo:
CacheManager = holds names of caches (pidname, qid)
CacheEvictionService = every 12 hours, evicts caches in cache manager

Will need to update deployment file with eviction frequency (43200000 ms)

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/CacheEvictionService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/CacheEvictionService.java
@@ -1,0 +1,23 @@
+package eu.dissco.core.handlemanager.service;
+
+import java.util.Objects;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.cache.CacheManager;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class CacheEvictionService {
+
+  private final CacheManager cacheManager;
+
+  @Scheduled(fixedRateString = "${caching.spring.ttl}")
+  public void evictAllCaches() {
+    log.info("Evicting all caches");
+    cacheManager.getCacheNames()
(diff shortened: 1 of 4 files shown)


src/main/java/eu/dissco/core/handlemanager/service/CacheEvictionService.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+@Slf4j
+@RequiredArgsConstructor
+public class CacheEvictionService {
+
+  private final CacheManager cacheManager;
+
+  @Scheduled(fixedRateString = "${caching.spring.ttl}")
```
southeo, line 17: will need to update deployment file...

src/main/java/eu/dissco/core/handlemanager/service/CacheEvictionService.java
```diff
@@ -0,0 +1,23 @@
...
+  public void evictAllCaches() {
+    log.info("Evicting all caches");
+    cacheManager.getCacheNames()
```
samleeflang, line 20: I thought there would be an easier way to do this, but look like you need the scheduler.
Also saw this example
https://www.baeldung.com/spring-setting-ttl-value-cache

samleeflang (approved):
We don't need this for ROR as well?

southeo:
> We don't need this for ROR as well?

We use the same method to resolve PID types and rors: 

```
@Cacheable("pidName")
public String getObjectName(String pid) {...}
```

Both RORs and FDO types have the field "name" in the same place! Lucky for us. 

https://api.ror.org/organizations/0566bfb96
https://hdl.handle.net/21.T11148/d8de0819e144e4096645

src/main/java/eu/dissco/core/handlemanager/service/CacheEvictionService.java
southeo: Yeah that's what I used to write this. Is there something from this example you're not seeing?

PR #81, opened 2024-04-16
Bug fix/update pid resolution issue

southeo:
Better logging, a quick filter issue, and check for a wikidata id properly


samleeflang (approved):
:spaghetti:

PR #29, opened 2023-06-16
Feature/upsert endpoint

southeo:
Introduces upsert endpoint that can create and update handles. Endpoint is PATCH `/api/v1/pids/upsert`

Checks if a handle for a given digital specimen already exists by checking primarySpecimenObjectId. Splits request into two components: new and updates. For the specimens to update, the issueNumber is incremented. 

This also includes code from the approved PRs "better archiving" and "rollback endpoints."

PR #83, opened 2024-04-29
Hardcode fdo profiles

southeo:
Breaking changes! Need to see following PRs:
- [Specimen Processor](https://github.com/DiSSCo/dissco-core-digital-specimen-processor/pull/48)
- [Annotation Processor](https://github.com/DiSSCo/dissco-core-annotation-processing-service/pull/43)
- [Media Processor](https://github.com/DiSSCo/dissco-core-digital-media-object-processor/pull/29)
- [Orchestration Service](https://github.com/DiSSCo/dissco-core-orchestration-backend/pull/41)

Purpose of this PR: to hardcode the names of the digital object types (PIDs) so we don't have to resolve them. And then some refactoring

## Properties file or enum?
Originally, we discussed putting these type names and pids in a properties file that could be overwritten. Instead, I put the fields in a hardcoded enum because:
1. By having an enum, we can more easily couple type names and pids in one object. e.g. FdoType.ANNOTATION contains the name "Annotation" and its pid in one object. Handy.
2. Injecting a properties file in the enum was messy
3. Overwriting the type names/pids in `application.properties` might not be a good idea. if these object types' names or pids change, we shouldn't have the old values in the code at all. 

### Why do some of the digital object names have capitals or spaces? 
Yeah that's the name of the object when you resolve it, so I went with that. It's not a consistent pattern, but it seems to be unchangeable. Hopefully we can fix this when we move to production epic. 

## Changes to the request objects
Here's what the requests used to look like: 

```
{
  "data": {
      "type":"digitalSpecimen", 
      "attributes": { 
         "digitalObjectType":"https://dtr-test.pidconsortium.net/objects/21.T11148/894b1e6cad57e921764e",
         "fdoProfile":"same as above..."
         "other attribute": "..."
     }
  }
}
```
It's redundant to have "type" on the higher level, then "digitalObjectType" in the request attributes. Now "type" will only accept the PID of the request type, and the field "digitalObjectType" is no longer accepted in the attribute body. 

I also removed the "fdoProfile" attribute from the request. This is because the profile is determined by the Type. The processing services/other clients shouldn't need to provide this redundant info. Currently, the same PID is used for profiles and Types. We don't have separate pids for these yet. I will make separate entries in the type registry after we discuss with ePIC. 

Obviously, these are breaking changes, and the processors are updated accordingly.

Changed request now looks like: 

```
{
  "data": {
      "type":"https://hdl.handle.net/10.500....", 
      "attributes": { 
         "digitalObjectType":"https://dtr-test.pidconsortium.net/objects/21.T11148/894b1e6cad57e921764e",
         "other attribute": "..."
     }
  }
}
```

## Other changes
- Refactored the tests a bit.

- "referentType" in DOIs and its inheritors is still poorly defined. The definition I have right now is: 

> A generic name for the type of object that the DOI refers to. This is different from digitalObjectType that points to a specific type, e.g. there can be different types of digital specimens that each have a slightly different metadata schema because they describe a different kind of specimen, like a botanical versus a geological specimen.

Anyway, there's no mapping for that term, so we just use digitalObjectName right now. 

- removed unused classes and redundant functions
- aligned media object fdo profile with what is in opends


src/main/java/eu/dissco/core/handlemanager/domain/fdo/vocabulary/specimen/TopicDiscipline.java
```diff
@@ -0,0 +1,147 @@
...
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.SPONGES;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.VERT_FOSSILS;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.VIRUSES;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Collections;
+import java.util.Set;
+import lombok.Getter;
+
+public enum TopicDiscipline {
```
samleeflang, line 65: Just to check, we don't calculate this right? But we take it over from the request

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -1,94 +1,94 @@
...
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_MEDIA_ID;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_MO_ID_NAME;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_MO_ID_TYPE;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_MO_TYPE;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_REFERENT_TYPE;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_SPECIMEN_OBJECT_ID;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_SPECIMEN_OBJECT_ID_NAME;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.PRIMARY_SPECIMEN_OBJECT_ID_TYPE;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.REFERENT_DOI_NAME;
+import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.REFERENT_NAME;
```
samleeflang, line 54: This could also be collapsed to `import static eu.dissco.core.handlemanager.domain.fdo.FdoProfile.*;`

samleeflang (commented):
Looks good, no big things, just a couple of questions.

src/main/java/eu/dissco/core/handlemanager/domain/fdo/FdoType.java
```diff
@@ -0,0 +1,76 @@
+package eu.dissco.core.handlemanager.domain.fdo;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@Getter
+public enum FdoType {
```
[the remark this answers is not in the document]
southeo: Would `FdoTypeInformation` then be an enum based on a properties file? I think that would still be ugly...
 
https://stackoverflow.com/questions/4908973/java-property-file-as-enum

made a user story to look at this further, as i'd like to push this through now: https://naturalis.atlassian.net/browse/DD-1163

src/main/java/eu/dissco/core/handlemanager/domain/fdo/vocabulary/specimen/TopicDiscipline.java
```diff
@@ -0,0 +1,147 @@
...
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.MaterialSampleType.ORG_PRODUCT;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.MaterialSampleType.OTHER_SOLID;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.MaterialSampleType.SLURRY_BIOME;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.MaterialSampleType.WHOLE_ORG;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.ALGAE;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.AMPHIBIANS;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.ANIMAL_GENE;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.ARACHNIDS;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.BACTERIA;
+import static eu.dissco.core.handlemanager.domain.fdo.vocabulary.specimen.TopicCategory.BIRDS;
```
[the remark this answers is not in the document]
southeo: the "optimize imports" function in intellij expands this, but I condensed it for now

src/main/java/eu/dissco/core/handlemanager/domain/fdo/vocabulary/specimen/TopicDiscipline.java
southeo: ya exactly

PR #85, opened 2024-05-02
Clean up exceptions

southeo:
- Exceptions can be made into runtime exceptions, as we just let them bubble up to the RestResponseEntityExceptionHandler
- Remove redundant exceptions
- Clean up the HandleNameGenerator and its tests

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
@@ -1,19 +1,19 @@
 package eu.dissco.core.handlemanager.service;
 
+import eu.dissco.core.handlemanager.exceptions.InvalidRequestException;
 import eu.dissco.core.handlemanager.properties.ApplicationProperties;
 import eu.dissco.core.handlemanager.repository.PidRepository;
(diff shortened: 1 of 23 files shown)


src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -30,115 +30,78 @@ public class PidNameGeneratorService {
...
     if (numberOfHandles < 1) {
-      return new ArrayList<>();
+      return Collections.emptySet();
     }
     if (numberOfHandles > applicationProperties.getMaxHandles()) {
-      log.warn("Max number of handles exceeded. Generating maximum {} handles instead",
-          applicationProperties.getMaxHandles());
-      numberOfHandles = applicationProperties.getMaxHandles();
+      log.error("Max number of handles exceeded : {} requested", numberOfHandles);
+      throw new InvalidRequestException("Max number of handles exceeded");
```
southeo, line 81: Rationale: if the specimen processor requests 5000 and only gets 1000 back, somewhere down the pipeline an unxpected exception (probably an NPE) will get thrown. Better to inform the client its request isn't being fulfilled.

PR #86, opened 2024-05-21
refactor exceptions

southeo:
apparently some files were being formatted according to the default intellij style instead of our sheet...that accounts for 85% of the changed lines sorry. 

1. Changes exceptions to runtime exceptions. We just let these exceptions bubble up to the RestResponseExceptionHandler anyway, so this cleans up code
2. refactors some tests, especially the FdoServiceTest, fixes some small issues
3. Renames PrimaryMediaObjectType to DcTermsType


samleeflang (commented):
Not a big fan of the exception changes. Could we keep them checked, if you want the methods to throw less, maybe we can make a super-exception for the custom exceptions? Normally you only make runtime exceptions when there is no way of recovering. In this case we can recover as we catch them in the restResponseEntityException and then throw a particular error.

samleeflang (approved):
:small_airplane:

PR #87, opened 2024-06-04
Bug fix/update serialization

southeo:
1. Adds "PID" element in update event to datacite
2. Serializes json node elements better using `TypeReference<Map<String, BaseJsonNode>>` instead of `TypeReference<Map<String, Object>>`

This is what I want to push to acceptance. Will port the fixes to main once this is approved. Coverage is a bit low but will fix that when i update main


samleeflang (approved):
:flying_saucer:

samleeflang (approved):

PR #88, opened 2024-06-20
MediaObject -> DigitalMedia

southeo:
Media Rename
- renames fdo type MediaObject to DigitalMedia
- Renames functions
- No impact on any other service (incl media processor)

Mapping rename
- mapping -> DataMapping
- Functions renamed
- no impact on orchestration service. If the location of the mappings moves from orchestration,dissco.tech/mapping/{id}, then we need to update the api to include the correct location in the pid record


samleeflang (approved):
Thanks for the update! Could you also update the openDS repo, haven't touched the fdo-type files yet.

PR #89, opened 2024-07-15
Mongo DB Storage

southeo:
## Data Storage
We've changed from a postgres rdb to a mongo doc store! :tada: 

Each pid record is a single document, identified by its handle. Each field in the pid record needs some metadata, which is now represented in a json:
- index
- type
- data
     - for regular fields: format (always "string"), and value
     - for HS_ADMIN: format="admin", index=200, value=0.NA/{prefix}, permissions = 011111110011
- ttl
- timestamp

An example document is found below:
```
{
  "_id": "20.5000.1025/ABC",
  "normalisedSpecimenObjectId": "06211:BIRD-COLLECTION-1:RMS-123",
  "values": [
    {
      "index": 1,
      "type": "digitalObjectType",
      "data": {
        "format": "string",
        "value": "https://hdl.handle.net/21.T11148/532ce6796e2828dd2be6"
      },
      "ttl": 86400,
      "timestamp": "2024-06-28 04:32:04Z"
    },
    {
      "index": 203,
      "type": "normalisedPrimarySpecimenObjectId",
      "data": {
        "format": "string",
        "value": "06211:BIRD-COLLECTION-1:RMS-123"
      },
      "ttl": 86400,
      "timestamp": "2024-06-28 04:32:04Z"
    }
  ]
}
```

### Indexing Local Identifiers
3 kinds of objects have local identifiers we can use to check if a handle has already been created for that kind of objects:
- Annotations: annotationHash
- specimens: normalisedSpecimenId
- Media: mediaUrl (called primaryMediaId in the handle API)

When we create a specimen, we first search on normalisedSpecimenId to see if a handle has already been created for that object. By putting normalisedSpecimenId on the top-level of the mongo document, we can index the field and quickly find if the handle exists.

We don't do this for annotations or media, but I've included the functionality, if we want to make the check.  

## Changes to Updates
We no longer support partial updates. Now, the request must contain all fields required to build a new 
- We keep a few fields from the previous version: pidIssueDate, fdoType, license, PID, PidStatus. These fields will have their original timestamp
- We increment pidIssueNumber
- We replace all other fields, even if they are unchanged, updating the timestamp

## Changes to Tombstone
Tombstoned records are not truncated anymore. Now, they have all the fields they had before they were tombstoned, but with additional tombstone metadata. 

## JsonSchemaValidator
- When we allowed partial updates, we had separate validation protocols for updates and creates because we didn't require the user to provide all the fields for an update. Now, all fields are required for both kinds of requests, so the validator got simplified
- Pretty sure we can remove this class honestly. We can the validator hassle by just accepting a specific object instead of a raw json. trying to keep this PR within its original scope though

## Other
- Cleaned up testsUtils
- Refactored some names


src/main/java/eu/dissco/core/handlemanager/repository/MongoRepository.java
samleeflang, line 1: Looks very neat this!

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -155,510 +167,620 @@ public class FdoRecordService {
...
+        String ui = applicationProperties.getUiUrl() + "/ds/" + handle;
+        return List.of(api, ui);
+      }
+      case DATA_MAPPING -> {
+        return List.of(applicationProperties.getOrchestrationUrl() + "/mapping/" + handle);
+      }
+      case SOURCE_SYSTEM -> {
+        return List.of(applicationProperties.getOrchestrationUrl() + "/source-system/" + handle);
+      }
+      case DIGITAL_MEDIA -> {
```
samleeflang, line 742: We could add the accesURI to this, do we have a task for that?

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -52,93 +47,61 @@ private Set<ByteBuffer> genHandleHash(int h) {
...
-  }
-
-  private String newSuffix() {
-    for (int idx = 0; idx < buf.length; ++idx) {
-      if (idx == 3 || idx == 7) { //
-        buf[idx] = '-'; // Sneak a lil dash in the middle
-      } else {
-        buf[idx] = symbols[random.nextInt(symbols.length)];
+    // We'll use this to make sure we're not duplicating results
+    HashSet<String> handleHash = new HashSet<>();
```
samleeflang, line 79: Why not just return the set?

src/main/java/eu/dissco/core/handlemanager/service/PidService.java
```diff
@@ -266,173 +271,170 @@ protected FdoType getObjectTypeFromJsonNode(List<JsonNode> requests) {
...
-      JsonNode requestAttributes = data.get(NODE_ATTRIBUTES);
-      FdoType type = FdoType.fromString(data.get(NODE_TYPE).asText());
-      var attributes = fdoRecordService.prepareUpdateAttributes(handle, requestAttributes, type);
-      attributesToUpdate.add(attributes);
+  private void verifyObjectsAreNew(List<String> normalisedIds)
+      throws InvalidRequestException, JsonProcessingException {
+    var existingHandles = mongoRepository
+        .searchByPrimaryLocalId(NORMALISED_SPECIMEN_OBJECT_ID.get(), normalisedIds);
+    if (!existingHandles.isEmpty()) {
+      log.error("Unable to create new handles, as ");
```
samleeflang, line 356: log message doesn't look complete

src/test/java/eu/dissco/core/handlemanager/repository/MongoRepositoryIT.java
```diff
@@ -0,0 +1,251 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+@Testcontainers
+class MongoRepositoryIT {
+
+  private static final String SPECIMEN_ID = PREFIX + "/SPECIMEN";
+  private static final String MEDIA_ID = PREFIX + "/MEDIA";
+
+  private static final DockerImageName MONGODB =
+      DockerImageName.parse("mongo:6.0.4");
```
samleeflang, line 45: I think 7.0 is out as well

samleeflang (commented):
How is the performance, do you feel it is equal to the db?
Really like how easy the mongodb repo looks

src/main/java/eu/dissco/core/handlemanager/configuration/FdoAttributeDeserializer.java
```diff
@@ -0,0 +1,19 @@
+package eu.dissco.core.handlemanager.configuration;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import eu.dissco.core.handlemanager.domain.repsitoryobjects.FdoAttribute;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class FdoAttributeDeserializer extends JsonDeserializer<FdoAttribute> {
```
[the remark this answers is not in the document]
southeo: should be deleted

src/main/java/eu/dissco/core/handlemanager/domain/repsitoryobjects/HandleData.java
```diff
@@ -0,0 +1,5 @@
+package eu.dissco.core.handlemanager.domain.repsitoryobjects;
+
+public interface HandleData {
```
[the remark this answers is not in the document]
southeo: Yeah so we have AdminHandleData and StringHandleData which implement this interface.

src/main/java/eu/dissco/core/handlemanager/repository/MongoRepository.java
southeo: :smile:

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
southeo: we already add the access URI, that what the primaryMediaId is.
See [here](https://github.com/DiSSCo/dissco-core-digital-media-object-processor/blob/9ac4b2759757083c2a93ba59d6999e6a599fac99/src/main/java/eu/dissco/core/digitalmediaprocessor/service/FdoRecordService.java#L68) in the media processor

southeo: fdo fields have diverged from ods... need to make time to align them

southeo:
Haven't tested large scale, but small tests are performing well. While we can't use the psql batch insert function anymore, i don't think handle creation will be a bottleneck. We now only insert one document instead of up to 30 rows per specimen. Updates are a lot better too, since we had something like `update (ATTRIBUTE, where HANDLE = xyz and INDEX = 111)` for EACH field we wanted to update. Now we just replace the document on the _id. And adding the local ids to the top-level is really good. 
:partying_face:

PR #90, opened 2024-07-22
Feature/jsonschema 2 pojo

southeo:
- uses ABC-request-attribute.json schema to build pojos instead of manually request objects for each type
- specifies the objects for requests instead of raw json. (just for type/id/attributes upon receiving request)
- removes jsonschemavalidator since the above points makes it redundant


samleeflang (approved):
:parasol_on_ground: 
Looks great, especially like how much code you were able to remove.

PR #97, opened 2024-08-19
Tombstone datacite pid

southeo:
- Sends a tombstone message to data
- Relationship type `HasMetadata` was chosen as it seemed to be the most general. DataCite enforces a vocabulary on relationshipType, but we don't. It would be nice if we used their vocabulary for this field, but until we have a vocabulary ourselves, it has to be as general as possible


samleeflang (approved):
:penguin:

PR #99, opened 2024-09-09
Feature/update tombstone terms

southeo:
tombstone terms were out of sync with tombstone metadata
Technically this should be part of the 1.0.0 fdo profile update, but since we're working on tombstoning objects this sprint, it's good to have this aligned sooner than later


samleeflang (approved):
:tophat:

PR #103, opened 2024-09-25
Feature/reserve handles

southeo:
Adds optional `draft` parameter to a create request. If this is a draft request, we mark the pid status as DRAFT instead of ACTIVE. In the case of DOIs, we do not notify DataCite. 

Adds `activate` endpoint, which marks Pid status as ACTIVE, and notifies DataCite if the PID is a doi
- note: we don't check if the PID is a handle or DOI. In test, which version of the api a client sends the request to (`handle-manager` or `doi-manager`) doesn't matter. In acceptance or production, the repositories will be different, so if a client sends a handle to be activated by the the doi manager, the PID won't be in the right repository, it won't be found, and so datacite won't be notified. 
- If a user tries to upsert a draft pid without activating it first, we activate it


src/main/java/eu/dissco/core/handlemanager/service/PidService.java
```diff
@@ -216,9 +218,31 @@ public JsonApiWrapperWrite searchByPhysicalSpecimenId(String normalisedPhysicalI
...
+    try {
+      draftRecords = mongoRepository.getHandleRecords(handles);
+    } catch (JsonProcessingException e) {
+      log.error("Unable to read PID record", e);
+      throw new InvalidRequestException("Unable to read PID record");
+    }
+    var activeRecords = draftRecords.stream().map(
+        draft -> {
+          var attributes = new ArrayList<>(draft.attributes());
+          attributes.set(attributes.indexOf(getField(attributes, PID_STATUS)),
```
samleeflang, line 236: Do we also update the version + versionTimestamp?

src/test/java/eu/dissco/core/handlemanager/service/HandleServiceTest.java
```diff
@@ -178,17 +187,33 @@ void testCreateHandle() throws Exception {
...
-    var result = service.createRecords(List.of(request));
+    var result = service.createRecords(List.of(request), false);
 
     // Then
     assertThat(result).isEqualTo(expected);
   }
 
+  @Test
+  void testActivateHandle() throws Exception {
+    var draftAttributes = genHandleRecordAttributes(HANDLE, CREATED, FdoType.HANDLE);
```
samleeflang, line 203: Missing the `// Given`

samleeflang (commented):
Ok exciting changes, nice to see that the changes are relatively limited

src/main/java/eu/dissco/core/handlemanager/service/PidService.java
southeo: wow I can reply to this in the ide

southeo: good catch, fixed now

samleeflang (approved):
:beans:

PR #106, opened 2024-10-11
fdo attributes are map instead of list

southeo:
Change fdo record attribute list to a enumMap so we can find attributes of interest faster. 

There is still a List in the FdoRecord object so we can read from/write to the database. 

Before, if we wanted to change the fdo record (e.g. pid status update, increment version, etc), we'd have to do something like:

```
attributeList.set(attributeList.indexOf(getField(attributeList, TARGET_ATTRIBUTE)), new FdoAttribute());

 public static FdoAttribute getField(List<FdoAttribute> fdoAttributes, FdoProfile targetField) {
    for (var attribute : fdoAttributes) {
      if (attribute.getIndex() == targetField.index()) {
        return attribute;
      }
    }
   return null;
}
```
Now we can do:

```
attributeMap.replace(TARGET_ATTRIBUTE, new FdoAttribute());
```

So we're going from multiple instances of a O(n^2) complexity function to a O(1) function. :tada:


src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
```diff
@@ -131,14 +136,14 @@ public class FdoRecordService {
   private static final String JSON_ERROR_MSG = "Unable to parse json request";
   private static final String FDO_LICENSE_VALUE = "https://spdx.org/licenses/CC0-1.0.json";
   private static final String FDO_LICENSE_NAME_VALUE = "CC0 1.0 Universal";
-  public static final List<Integer> GENERATED_KEYS;
+  public static final Set<FdoProfile> GENERATED_KEYS;
```
samleeflang, line 139: Calling the object FdoProfile feels a bit strange. Why did you choose for this name?

samleeflang (approved):
Excellent, nice change! Just got a small question about the name

src/main/java/eu/dissco/core/handlemanager/service/FdoRecordService.java
southeo: Way back, like pre schemas.dissco.tech way back, I used this class to define as the source of truth as to what the FDO profiles contained. It can still sort of be thought that way, the enum class **as a whole** being a collection of terms that define what goes into a FDO record. Obviously, a better name for this class would be "FdoAttribute", but that's already taken. `FdoAttribute`s are objects that actually make up the mongodb record, and `FdoProfile`s define the term name and its corresponding index. 

Maybe `FdoTerm` makes more sense, but then I'd worry it's still too similar to `FdoAttribute`... 

I'm open to changing this, just not sure what to

PR #107, opened 2024-10-23
ingestion fixes

southeo:
- Fix bug with updates in fdo record. When retrieving fdo record for updates, we need to fill the map<FdoProfile, FdoAttribute> instead of returning an empty map. 
- Adds `FdoType.fromString()` function
- Logging improvements
- Doesn't fail if requests are empty


samleeflang (approved):
:train2:

PR #113, opened 2024-12-04
Update for the FDO types and the routes

samleeflang:
Small PR for some updates needed for production.
Small change to the route so it we can switch the endpoint
Small change in the FDO types to align with https://github.com/DiSSCo/dissco-developers-documentation/wiki/FDO-types-and-profiles


src/main/java/eu/dissco/core/handlemanager/domain/fdo/FdoType.java
```diff
@@ -32,35 +32,35 @@ public enum FdoType {
   DIGITAL_SPECIMEN(
```
southeo, line 32: You have to update the `@JsonProperty` and `@JsonAlias` or these values won't map to the correct enum

southeo (commented):
:pancakes:

PR #116, opened 2024-12-09
Feature/fdo update

southeo:
- align with fdo 1.0
- introduces "unrecognized" fdo property as a small workaround in the mongodb layer so the handle api won't crash if a pid record has a property from a previous version

### DIFF
--- a/src/main/resources/json-schema/digital-media-request-attributes.json
+++ b/src/main/resources/json-schema/digital-media-request-attributes.json
@@ -61,35 +61,44 @@
       "type": "string",
       "description": "Only mimetypes for audio, image, model, video, text should be provided. https://www.iana.org/assignments/media-types/media-types.xhtml"
     },
+    "license": {
+      "type": "string",
+      "description": "Maps to dcterms:license, unenforced value that may or may not be machine actionable. http://purl.org/dc/terms/license.",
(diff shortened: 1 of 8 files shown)


src/main/resources/json-schema/digital-media-request-attributes.json
```diff
@@ -61,35 +61,44 @@
...
   },
   "required": [
+    "referentName",
     "mediaHost",
     "mediaHostName",
     "linkedDigitalObjectPid",
     "linkedDigitalObjectType",
-    "primaryMediaId"
+    "primaryMediaId",
+    "license",
```
samleeflang, line 100: `license` is a required field but in the Service I don't think we actually set `license`. Couldn't that create a potential issue?

samleeflang (commented):
Comment about license

src/main/resources/json-schema/digital-media-request-attributes.json
southeo: Ok this should be removed, also made a [pr on opends](https://github.com/DiSSCo/openDS/pull/225). If we don't have a license in the FDO record I think we'll somehow survive

samleeflang (approved):
:card_index_dividers:

PR #119, opened 2025-01-07
Bug fix/normalize media

southeo:
see [jira](https://naturalis.atlassian.net/browse/DD-1587?atlOrigin=eyJpIjoiOGFhNGEzOGNmYTg2NGY5NzkzMzJiOWUwY2VkZmY4YmQiLCJwIjoiaiJ9) for more information

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/ServiceUtils.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/ServiceUtils.java
@@ -1,5 +1,6 @@
 package eu.dissco.core.handlemanager.service;
 
+import eu.dissco.core.handlemanager.schema.DigitalMediaRequestAttributes;
 import java.util.stream.Collector;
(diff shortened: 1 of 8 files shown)


src/main/java/eu/dissco/core/handlemanager/service/ServiceUtils.java
```diff
@@ -22,4 +23,8 @@ private ServiceUtils() {
     );
   }
 
+  public static String normalizeMediaId(DigitalMediaRequestAttributes request) {
+    return request.getLinkedDigitalObjectPid() + ":" + request.getPrimaryMediaId();
```
samleeflang, line 27: Ok this is probably good. Although with this particular issue this wouldn't have solved it as it were multiple images with the same URL for the same linkedDigitalObject. But we will fix that at ingestion so that should never occur again

samleeflang (approved):
:cake:

PR #124, opened 2025-02-17
Batch Auto Annotations - Reduce response for annotation requests

southeo:
Reduces response for created annotations. The annotation processor only needs either the hash or the target identifier. Because we're starting to be batching annotations in groups of ~300, we don't want each individual response to be too large.


samleeflang (approved):
:+1:

PR #127, opened 2025-04-22
Switch from Kafka to RabbitMQ

samleeflang:
Switch from Kafka to RabbitMQ

https://naturalis.atlassian.net/browse/DD-1679


src/main/java/eu/dissco/core/handlemanager/configuration/RabbitMqConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.AllArgsConstructor;
+import org.springframework.amqp.rabbit.config.SimpleRabbitListenerContainerFactory;
+import org.springframework.amqp.rabbit.connection.ConnectionFactory;
+import org.springframework.amqp.rabbit.core.RabbitTemplate;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@AllArgsConstructor
+public class RabbitMqConfiguration {
```
southeo, line 13: maybe specify for the DOI profile only? we only send messages to datacite in that profile

southeo (approved):
:dodo:

src/main/java/eu/dissco/core/handlemanager/configuration/RabbitMqConfiguration.java
samleeflang: Added

PR #128, opened 2025-05-02
digital media key -> primary media id

southeo:
Removes reference of linked object (i.e. the specimen) in media fdo profile and handle response

[Jira](https://naturalis.atlassian.net/browse/DD-1644)


samleeflang (approved):
Nice that they can now use the same method. :trophy:

PR #133, opened 2025-06-11
filter on active handles

southeo:
- Filter on active handles when retrieving by local id
- Change a list which would be better as a map


samleeflang (approved):
:cactus:

PR #134, opened 2025-06-24
Java 21

southeo:
- Update to java 21
- Remove unused ServiceUtils class
- New way of constructing URL objects -- previous is deprecated


samleeflang (approved):
Looks good :ear_of_rice:

PR #137, opened 2025-09-04
Manual PIDs

southeo:
**PID List**
Uses a file defined in applications.properties to produce new PIDs (either handles or dois). 
If we run out of PIDs, we start generating PIDs randomly. 

We ignore any PID that doesn't include the correct prefix (set in application.properties)

**Overwrite**
If the overwrite property is set, we overwrite any previous record that exists at that PID. 

We if this is set, we have to make sure to send the right kind of event to datacite (create or update), so there's an extra db call. (we normally just check the accessURI or physical Specimen ID to see if the object is ingested)

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
+++ b/src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
@@ -5,19 +5,20 @@ import static eu.dissco.core.handlemanager.properties.ProfileProperties.HANDLE_D
 
 import eu.dissco.core.handlemanager.properties.ApplicationProperties;
 import eu.dissco.core.handlemanager.repository.MongoRepository;
-import java.io.BufferedReader;
 import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
@@ -36,6 +37,7 @@ public class PidNameGeneratorService {
   private final char[] buf = new char[LENGTH];
   private final MongoRepository mongoRepository;
   private final Random random;
+  private static final int BATCH_SIZE = 10000;
 
 
   public Set<String> generateNewHandles(int h) {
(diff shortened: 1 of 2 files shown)


src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -34,9 +44,68 @@ public Set<String> generateNewHandles(int h) {
...
+      log.info("Manual pids are depleted. Generating {} random pids", h - pids.size());
+      pids.addAll(genHandleHashSet(h - pids.size()));
+      applicationProperties.setManualPidFile(null);
+    }
+    return pids;
+  }
+
+  private void overwriteManualPidFile() {
+    var oldFile = new File(applicationProperties.getManualPidFile());
+    var writeFile = (new File(getWriteFileName()));
```
samleeflang, line 88: extra paranthesis can be removed

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -34,9 +44,68 @@ public Set<String> generateNewHandles(int h) {
...
+      }
+    } catch (IOException e) {
+      log.error("Error while reading manual pid file. Generating random PIDs", e);
+      return genHandleHashSet(h);
+    }
+    overwriteManualPidFile();
+    if (pids.size() < h) {
+      log.info("Manual pids are depleted. Generating {} random pids", h - pids.size());
+      pids.addAll(genHandleHashSet(h - pids.size()));
+      applicationProperties.setManualPidFile(null);
```
samleeflang, line 81: At the restart this null will be overwritten, could that be an issue?

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -34,9 +44,68 @@ public Set<String> generateNewHandles(int h) {
           applicationProperties.getMaxHandles());
       h = applicationProperties.getMaxHandles();
     }
+    if (applicationProperties.getManualPidFile() != null) {
+      return generateManualPids(h);
+    }
     return genHandleHashSet(h);
   }
 
+  private Set<String> generateManualPids(int h) {
```
samleeflang, line 53: Ok I wonder if the ouput input file is the easiest/best way forward as you will be recreating potentially large file (400.000+ line?).  You could probably do the else here (so we got all we need, write to a file from here) more efficiently.  Although it might not matter too much

src/test/java/eu/dissco/core/handlemanager/service/PidNameGeneratorServiceTest.java
```diff
@@ -56,6 +74,94 @@ void testSingleBatchGen() {
...
+
+    // When
+    var result = pidNameGeneratorService.generateNewHandles(4);
+
+    // Then
+    assertThat(result).isEqualTo(expected);
+    assertThat(fileHasExpectedSize(0, path)).isTrue();
+  }
+
+  private static boolean fileHasExpectedSize(int size, Path path) throws IOException {
```
samleeflang, line 144: Normally static methods go to the top

samleeflang (commented):
Ok see what you done here. I might have done the file tricks differently but I think it is ok. Check if it is fast enough if you use a massive file.

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
southeo: I don't think so. If we restart, and the file name is reinstated, then the service will immediately see there's no pid names in the file and mark it as null. I don't love it, but it should hold up. 

We won't be running the application in this mode very often.

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
samleeflang: Would need to dig into it but if it works, let's keep it like this for now

PR #139, opened 2025-09-10
Feature/manual pids from db

southeo:
Uses db table instead of reading/writing to a file for manual pids
Note: Reduces the number of application properties
Note: I've already reverted the previous solution, where I tried with a file instead. 
Note: Line count is misleading! 300 lines are from jooq, 300 lines are from  testing! This is straightforward, I promise.


pom.xml
```diff
@@ -227,6 +240,11 @@
     <plugins>
       <plugin>
         <artifactId>jacoco-maven-plugin</artifactId>
+        <configuration>
+          <excludes>
+            <exclude>eu.dissco.core.handlemanager.jooqobjects</exclude>
```
samleeflang, line 245: Why do we exclude this here?

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
```diff
@@ -34,9 +37,23 @@ public Set<String> generateNewHandles(int h) {
...
   }
 
+  private Set<String> generateManualPids(int h) {
+    var pids = new HashSet<>(manualPidRepository.getPids(h));
+    var manualPids = new HashSet<>(pids);
+    while (pids.size() < h) {
+      log.warn("Manual pids depleted. Generating manual pids");
+      pids.addAll(genHandleHashSet(h - pids.size()));
+    }
+    manualPidRepository.deleteTakenPids(manualPids);
```
samleeflang, line 53: What if it fails after this? Shouldn't the delete take place when we are certain the PID was successfully used.

src/test/java/eu/dissco/core/handlemanager/repository/BaseRepositoryIT.java
```diff
@@ -0,0 +1,44 @@
...
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+
+@Testcontainers
+class BaseRepositoryIT {
+
+  private static final DockerImageName POSTGIS =
+      DockerImageName.parse("postgres:15.5").asCompatibleSubstituteFor(IMAGE);
```
samleeflang, line 21: Update this to 17.5 or more recent

samleeflang (commented):
Couple small comments

src/main/java/eu/dissco/core/handlemanager/service/PidNameGeneratorService.java
southeo: Very good point. 
I was worried that we only want to delete the manual pids, not the randomly generated pids. What if the randomly generated pids happen to be somewhere in the database? 

The solution: I think that manual pids need to always already be in the database (maybe in our case, but also maybe just as "reserved"). That way, the pid name generator service won't return a pid that's in the manual pid table, and we won't delete anything we shouldn't.

pom.xml
southeo: Was trying to make sonar happy by excluding the new jooq stuff from code coverage, doesn't seem to work though. Will remove

samleeflang: Yeah it is in the github action where you can exclude stuff from sonar

southeo: ty!

southeo: wow sonar is so happy :)

samleeflang (approved):
Relational DB for the win :relaxed:

PR #140, opened 2025-12-03
Add role protection to endpoint

samleeflang:
Addded dissco-handle-manager and role to Keycloak.
Will give this role to all application which are creating PID's

(let's see if this automatically links the PR)

### DIFF
--- a/src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
+++ b/src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
@@ -46,6 +46,7 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.security.core.Authentication;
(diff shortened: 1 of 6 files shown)


src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
```diff
@@ -150,6 +151,7 @@ public ResponseEntity<JsonApiWrapperWrite> searchByPrimarySpecimenObjectId(
       }),
       @ApiResponse(responseCode = "204", description = "Received empty request")
   })
+  @PreAuthorize("hasRole('dissco-handle-manager')")
```
southeo, line 154: ahh i'm just wondering if we should have separate roles for doi and handle managers? I think it's a different subset of agents who will be making handles and dois. 
handles: likely just internal applications
dois: trusted parties (because dois cost us money)

this might be tricky to do with the shared controller tho :///

southeo (commented):
:key:

src/main/java/eu/dissco/core/handlemanager/controller/PidController.java
samleeflang: We discussed this right? It might be tricky, for now this should provide enough security.

southeo (approved):
🥼

PR #141, opened 2025-12-05
add upsert for pids

southeo:
If we are using manual PIDs, and we run out of manual PIDs, we need to do a mongodb upsert, not update. If we have a mix of manual pids and randomly generated ones, then some records will already exist, but others will not. further calls to the api when isUseManualPids = true when there are no more pids in the db will result in all new pids being generated but not inserted into the db.

### DIFF
--- a/src/test/java/eu/dissco/core/handlemanager/repository/MongoRepositoryIT.java
+++ b/src/test/java/eu/dissco/core/handlemanager/repository/MongoRepositoryIT.java
@@ -239,6 +239,36 @@ class MongoRepositoryIT {
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


src/test/java/eu/dissco/core/handlemanager/repository/MongoRepositoryIT.java
samleeflang, line 1: I think a test to check the upsert functionality might be missing

samleeflang (approved):
Approving but I think I am missing a test for the mongoDB upsert functionality. If it is there just merge

